PR #66, opened 2015-06-18
Do not raise a sub-context's error

ramontayag:
Happens when the an organizer-action rescues errors and a child action raises an exception before being able to set a promised key.

Fix is to not rely on the context's `current_action`, which is overwritten every time an action is executed.

### DIFF
--- a/lib/light-service/context/key_verifier.rb
+++ b/lib/light-service/context/key_verifier.rb
@@ -2,9 +2,9 @@ module LightService; class Context
   class KeyVerifier
     attr_reader :context, :action
 
-    def initialize(context)
+    def initialize(context, action)
       @context = context
-      @action = context.current_action
+      @action = action
     end
 
     def are_all_keys_in_context?(keys)
@@ -40,13 +40,13 @@ module LightService; class Context
       context
     end
 
-    def self.verify_keys(context, &block)
-      ReservedKeysVerifier.new(context).verify
-      ExpectedKeyVerifier.new(context).verify
+    def self.verify_keys(context, action, &block)
+      ReservedKeysVerifier.new(context, action).verify
+      ExpectedKeyVerifier.new(context, action).verify
 
       block.call
 
-      PromisedKeyVerifier.new(context).verify
+      PromisedKeyVerifier.new(context, action).verify
     end
   end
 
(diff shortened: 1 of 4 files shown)


adomokos:
Thank you for your PR @ramontayag! I'll check it out this weekend.

lib/light-service/context/key_verifier.rb
```diff
@@ -2,9 +2,9 @@ module LightService; class Context
   class KeyVerifier
     attr_reader :context, :action
 
-    def initialize(context)
+    def initialize(context, action)
       @context = context
-      @action = context.current_action
+      @action = action
```
adomokos: Could we remove `context.current_action`?

ramontayag: I suppose so. I wasn't sure if it was used anywhere else (I didn't check though).

lib/light-service/context/key_verifier.rb
```diff
@@ -40,13 +40,13 @@ def verify
...
-      ExpectedKeyVerifier.new(context).verify
+    def self.verify_keys(context, action, &block)
+      ReservedKeysVerifier.new(context, action).verify
+      ExpectedKeyVerifier.new(context, action).verify
 
       block.call
 
-      PromisedKeyVerifier.new(context).verify
+      PromisedKeyVerifier.new(context, action).verify
     end
```
adomokos: @ramontayag: I ran all the specs and they pass in your PR. However, I am trying to understand why we want to do this before I merge the code in. Here is what puzzles me.

I added this to line 44:
```ruby
pp context.current_action == action
```

This evaluates to true. Then I changed line 45 to `ReservedKeysVerifier.new(context, context.current_action).verify`. I changed the same for ExpectedKeyVerifier and PromisedKeyVerifier initializer. When I run the tests, I am getting the PromisedKeyNotFoundError.

Do you know why? Can you explain?

ramontayag: @adomokos if I understand you right:

When you change the verifiers to accept `context.current_action` as the second argument, then all the actions share a state. This has the same effect as just passing `context` and extracting the `current_action` from the context. Whenever there's a change to `context.current_action`, the verifiers become affected.

This pull request keeps that state away from the key verification process. Yes, `context` still needs to be passed because that's where the keys live, but the state remains unchanged during the key verification of that action.

I hope I answered your question.

adomokos: You did. Thank you!

Issue #67, opened 2015-06-18
Allowing aliases for :expects attributes on a LightService::Context object


rewinfrey:
Let's say we have a generic action that expects a `user` to be supplied in the context.

Later, we create a new Organizer and for the context of that Organizer and its associated Actions, instead of `user` we care about the idea of an `owner`.

However, I want to use the generic action that expects the context to have an `user`. Currently, this causes me to have to create a small, specific action to convert the idea of an `owner` to an `user`:

```ruby
class ExampleAction
  expects :owner
  promises :user

  executed do |context|
    context.user = context.owner
  end
end
```

This can quickly get out of hand though. Alternatively, at the time I create the `owner` object, I could associate it with an `user` at the same time:

```ruby
class ExampleAction2
  expects :params
  promises :owner, :user

  executed do |context|
    context.user = context.owner = User.new(context.params)
  end
end
```

The trouble I have here is that our use of an action in this Organizer requiring an `user` leaks into an action that is really focused on the idea of an `owner`, which means that a more generic action meant to be reused is now leaking into other actions in order to be reused.

I'd like to propose we add the idea of expects aliases to LightService:

```ruby
class ExampleAction3
  expects :user
  alias_context :user, :owner

  executed do |context|
    context.user.behavior!

    # which is equivalent to saying:
    # context.owner.behavior!
  end
end
```

What I hope `ExampleAction3` demonstrates is how LightService can alias an object in the context
with another object in the context, and correctly send messages to the aliased object in the context when the aliased object exists. If there is no aliased object in the context, then `expects` and the context validation works as normal.

adomokos:
:+1: for this feature!

jpmoral:
In my opinion, having the alias in `ExampleAction3` also leaks information that somewhere, some other action passes an `owner` instead of a `user`.  It seems that `ExampleAction3` shouldn't have to know about the other possible names of what it expects.

What do you think of putting such functionality in the organizer?  If we think of actions as self-contained parts, it seems that the organizer should be responsible for making the parts fit together.

jpmoral:
@rewinfrey @adomokos 
Please see my pull request: #68

adomokos:
Thanks @jpmoral! Closing this Issue.

PR #68, opened 2015-07-04
Context key aliases

jpmoral:
This addresses #67.  An `aliases` macro was added to the Organizer module.  It accepts a hash; the keys are the original keys and the values are the key aliases.

This is passed to `WithReducer#with`.  `WithReducer#reduce` inserts new key-value pairs into the context: the keys are the key aliases while the values are the values corresponding to the original keys.  Thus each action can use the aliases when accessing the context.  Example usage:

```ruby
class AnOrganizer

  aliases my_key: :key_alias

  def self.for_order(order)
    with(:order => order).reduce(
        AnAction,
        AnotherAction,
      )
  end
end

class AnAction
  extend LightService::Action
  promises :my_key

  executed do |context|
    context.my_key = "value"
  end
end

class AnotherAction
  extend LightService::Action
  expects :key_alias

  executed do |context|
    context.key_alias # => "value"
  end
end
```

See `spec/organizer_key_aliases_spec.rb` for more.

### DIFF
--- a/README.md
+++ b/README.md
@@ -200,6 +200,46 @@ end
 
 Take a look at [this spec](spec/action_expects_and_promises_spec.rb) to see the refactoring in action.
 
+## Key Aliases
+The `aliases` macro sets up pairs of keys and aliases in an Organizer.  Actions can access the context using the aliases.
+
+This allows you to put together existing Actions from different sources and have them work together without having to modify their code.  Aliases will work with or without Action `expects`.
+
+Say for example you have actions `AnAction` and `AnotherAction` that you've used in previous projects.  `AnAction` provides `:my_key` but `AnotherAction` needs to use that value but expects `:key_alias`.  You can use them together in an organizer like so:
+
+```ruby
+class AnOrganizer
+  extend LightService::Organizer
+
+  aliases my_key: :key_alias
+
+  def self.for_order(order)
+    with(:order => order).reduce(
+        AnAction,
+        AnotherAction,
+      )
+  end
+end
+
+class AnAction
+  extend LightService::Action
+  promises :my_key
+
+  executed do |context|
+    context.my_key = "value"
+  end
+end
+
+class FooAction
+  extend LightService::Action
+  expects :key_alias
+
+  executed do |context|
+    context.key_alias # => "value"
+  end
+end
+```
+
 ## Logging
 
 Enable LightService's logging to better understand what goes on within the series of actions,
--- a/lib/light-service/action.rb
+++ b/lib/light-service/action.rb
@@ -69,8 +69,6 @@ module LightService
       def all_keys
         expected_keys + promised_keys
       end
-
     end
-
   end
 end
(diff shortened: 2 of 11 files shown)


adomokos:
@jpmoral: Thank you for working on this PR!

You have many good changes here that I'd like to have in LS. I looked at how you are setting the aliases in the `WithReducer` class. I was wondering if we could delegate that responsibility to the `Context` class.

Check out [this commit](https://github.com/adomokos/light-service/commit/d4fad8b892ae6b9fb3a48a8938a938dc86488695). I tried to see how it would work, but the specs are not passing there.

What do you think?

jpmoral:
@adomokos 
I think it would be nice to have `Context` handle the aliases, and I did give it some thought while writing the code.
The main issue I faced was that we do not know beforehand when the keys to be aliased will be available in the context: whether the organizer or one of the actions provides them.  That's why I put the aliasing in `WithReducer`.  I admit `put_aliases_in_the_context` is an awkward name and is called in an awkward place.
I have two suggestions:

1.  [Call `Context#set_aliases` before each action execution](https://github.com/adomokos/light-service/compare/jpmoral-key-aliases...jpmoral:key-aliases-option-1?expand=1)

2.  Give `Context` the aliases and have it automagically return the correct values.  (Working on the code, will post a link to a diff later.)

jpmoral:
@adomokos 
Thoughts?

adomokos:
@jpmoral: these are great suggestions, thank you for thinking about it!!

[Here is](https://github.com/adomokos/light-service/commit/2827f550e5aa40c71a5c0732a8c7c1f8903a7bcf) what I hacked on tonight. It's not perfect as it still has one spec failing, but I think it clearly expresses what I'd like to see. Please note how I use the data to attach the `_aliases` key-value pair to it in the `Organizer` class. This way the data is a pass-through in `WithReducer`, that class is not aware of the `aliases` collection.

Now only the `Organizer` and the `Context` are the two entities or objects that are aware of the `aliases` hash. Could you get this one to work?

jpmoral:
@adomokos 
Please see the updated diff.  I liked how `WithReducer` no longer knows about the aliases.  If the changes are good there are a few more specs I'd like to add.  Also, do you mind if I squash the new commits together?  Some of them don't make sense as individual commits.

adomokos:
@jpmoral: I really like where you're going with this. Please squash the commits, add the specs and I'll merge in the code. Really nice work!! :+1:

README.md
```diff
@@ -200,6 +200,46 @@ end
...
+class AnAction
+  extend LightService::Action
+  promises :my_key
+
+  executed do |context|
+    context.my_key = "value"
+  end
+end
+
+class FooAction
```
adomokos: This should be `AnotherAction` in this example.

jpmoral: Good catch!

jpmoral:
Closing this in favor of #69.

PR #72, opened 2015-07-18
Assign expected and promised keys to Action subclass

jpmoral:
Currently when an `Action` is subclassed the subclass cannot access the expected and promised keys using dot notation.
```ruby
class AnAction
  extend LightService::Action
  expects :expected_key
  executed do |context|
    # do stuff
  end
end

class AnActionSubclass
  executed do |context|
    # do stuff
    context.expected_key # raises an error
  end
end
```

This PR will allow the above to work.

### DIFF
--- a/lib/light-service/action.rb
+++ b/lib/light-service/action.rb
@@ -10,6 +10,12 @@ module LightService
       base_class.extend Macros
     end
 
+    def inherited(subclass)
+      subclass.instance_variable_set(:@_expected_keys, @_expected_keys)
+      subclass.instance_variable_set(:@_promised_keys, @_promised_keys)
+      super(subclass)
+    end
+
     module Macros
       def expects(*args)
         @_expected_keys ||= []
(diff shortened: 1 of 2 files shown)


jpmoral:
@adomokos 

I discovered this a couple of weeks ago.  I was able to eliminate the need for subclassing by going with a different (cleaner) design of Action and Organizer, so I'm not sure how useful this is.

However, I made the PR because it might be surprising to others that `expects` and `promises` don't work with inheritance.

adomokos:
@jpmoral, thank you for submitting this PR.

The reason I opted to use class methods for the actions was to limit inheritance. With inheritance there is an "inferred" state that I tried to avoid.

Would you be OK if I did not merge this PR? I know you worked on it, that's why I feel bad about it.

jpmoral:
@adomokos

It's no problem, like I mentioned, even I'm a bit iffy about it.  It might be helpful though to be explicit in the docs that inheritance of actions isn't really encouraged.

adomokos:
Good suggestion!

PR #79, opened 2016-01-15
Adds around_each to WithReducers for organizers

bwvoss:
The ambition of this commit is to unobtrusively
gather arbitrary metrics from your application.

An around_each receiver follows the duck type of:

```ruby
def call(action, context)
```
The rest is up to you!

Here is an example that could be used to check persisted counts in a
data pipeline (there is another example in the README for logging
duration or profiling actions):

```ruby
class DataCounterCheck
  def self.call(action, context)
    before_count = context.data.count
    result = yield
    after_count = context.persisted_data.count

    LightService::Configuration.logger.info({
      :action       => action,
      :before_count => before_count,
      :after_count  => after_count,
      :difference   => before_count - after_count,
    })

    result
  end
end
```

### Design Consideration

I tried to focus most on the readability of the API, and I think it reads nicely method-chained.
I chose to simply yield the block passed by light-service so the around each objects do not have
to change even if the way light-service chooses to execute actions does.  It's fairly open-ended, but still simple enough, I hope.

### Complexity

The around_each method call provides enough to do the use cases I can
foresee without overcomplicating the software.  It does add some
assignments and conditionals, so it is close to a new abstraction, but
probably not quite yet.  It changes the flog by such:

```flog lib/light-service/organizer -g```

Before:

```
123.2: flog total
     7.7: flog/method average

    55.5: LightService::Organizer::WithReducerLogDecorator total
    40.2: LightService::Organizer::WithReducerLogDecorator#reduce
          lib/light-service/organizer/with_reducer_log_decorator.rb:27
    15.2: LightService::Organizer::WithReducerLogDecorator#with
          lib/light-service/organizer/with_reducer_log_decorator.rb:13

    19.0: LightService::Organizer::WithReducer total
    11.6: LightService::Organizer::WithReducer#reduce
          lib/light-service/organizer/with_reducer.rb:10
     7.4: LightService::Organizer::WithReducer#reduce_rollback
          lib/light-service/organizer/with_reducer.rb:28
```

After:

```
128.4: flog total
    7.6: flog/method average

    55.5: LightService::Organizer::WithReducerLogDecorator total
    40.2: LightService::Organizer::WithReducerLogDecorator#reduce
          lib/light-service/organizer/with_reducer_log_decorator.rb:27
    15.2: LightService::Organizer::WithReducerLogDecorator#with
          lib/light-service/organizer/with_reducer_log_decorator.rb:13

    23.1: LightService::Organizer::WithReducer total
    15.7: LightService::Organizer::WithReducer#reduce
          lib/light-service/organizer/with_reducer.rb:15
     7.4: LightService::Organizer::WithReducer#reduce_rollback
          lib/light-service/organizer/with_reducer.rb:37
```
### Benchmarks

It has a negligible impact on the organizer's reduction performance -- it is about 1.01x slower with around_each calls.

From this script https://gist.github.com/bwvoss/8483ecc4a0315e724fed, here are some before and after IPS benchmarks:

```
Calculating -------------------------------------

         default     884.000  i/100ms
         around_each 855.000  i/100ms

-------------------------------------------------

         default      8.912k (± 2.4%) i/s - 45.084k
         around_each  8.786k (± 3.6%) i/s - 44.460k

Comparison:

         default:     8911.9 i/s
         around_each: 8786.1 i/s - 1.01x slower
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,17 @@
 language: ruby
+
 rvm:
-  - 1.9.3
   - 2.0.0
   - 2.1.2
+  - 2.2.0
+  - 2.3.0
+
+before_install:
+  - gem install bundler
+
 # uncomment this line if your project needs to run something other than `rake`:
 script: bundle exec rspec spec
+
 gemfile:
   - gemfiles/activesupport_3.gemfile
-  - gemfiles/activesupport_4.gemfile
+  - gemfiles/activesupport_4.gemfileset
--- a/README.md
+++ b/README.md
@@ -146,6 +146,39 @@ end
 I gave a [talk at RailsConf 2013](http://www.adomokos.com/2013/06/simple-and-elegant-rails-code-with.html) on
 simple and elegant Rails code where I told the story of how LightService was extracted from the projects I had worked on.
 
+You can also add objects that get executed around every action, which is ideal for specialized logging:
+
+```ruby
+class LogDuration
+  def self.call(action, context)
+    start_time = Time.now
+    result = yield
+    duration = Time.now - start_time
+    LightService::Configuration.logger.info({
+      :action   => action,
+      :duration => duration
+    })
+
+    result
+  end
+end
+
+class CalculatesTax
+  extend LightService::Organizer
+
+  def self.for_order(order)
+    with(:order => order).around_each(LogDuration)reduce(
+        LooksUpTaxPercentageAction,
+        CalculatesOrderTaxAction,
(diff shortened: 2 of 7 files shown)


adomokos:
Thank you, @bwvoss!

rewinfrey:
:thumbsup: @bwvoss, I really like the analysis too!

bwvoss:
Thanks guys, it was a fun PR!

PR #83, opened 2016-04-13
Project page should update together with the README

padi:
While preparing for the lightning talk, I noticed that the website doesn't update together with the latest README.
Made flatdoc fetch from the latest README so that we don't have to update the github page manually.

preview: padi.github.io/light-service (images should be fixed when it gets merged to your gh-pages branch)


adomokos:
This is fantastic! Can you take a look at Travis? It seems to be unhappy.

jeroenvisser101:
Seems to be related to one of the dependencies requiring Node v4 or above, I think it's unrelated, but updating travis.yml to use newer node should solve that

padi:
Done 👍

adomokos:
Thank you! ❤️

adomokos:
@padi: It seems the images are not showing up properly there -

ls-image-off

Would you mind taking a look? This is the URL I am using: http://adomokos.github.io/light-service/

padi:
@adomokos Looks like the relative links won't work even if it's under your github page. I thought it didn't work before just because it's relative to my github page. It's now fixed.

PR #84, opened 2016-04-15
Make relative links absolute to fix project page

padi:
Should fix issues created by #83. I thought the images would work if it's relative to your github page. Looks like it's not.


adomokos:
Yeah, I wish we could have used relative image paths. But this works, thank you for it! 👍

PR #88, opened 2016-09-16
Slimming down ActiveSupport dependency

adomokos:
In order to get rid of ActiveSupport, I'd like to trim the project's
dependency on it. This is the first step of doing that.


adomokos:
Due to using I18n through ActiveSupport, getting rid of it might take more effort. Closing this for now, although it's good to see that AS dependency can be slimmed down.

PR #89, opened 2016-09-16
Require AS deprecation, that way we can upgrade to AS 5

### LINKED ISSUE #86: Add support for ActiveSupport 5
Right now, LS will break with AS5. I specified `"< 5.0"` for now, but we should make it work with and without 5.0.

adomokos:
I suspect auto-require of AS deprecation has been removed from
the project, even Rails [requires it](https://github.com/rails/rails/blob/master/activesupport/lib/active_support/core_ext/time/marshal.rb) in files it is using it.

Closes #86


jeroenvisser101:
Just one thought, we're now mixing single and double quotes, any rule here?

adomokos:
Good point, I would have expected Rubocop warning about that. All the requires in `lib/light_service.rb` are using single quotes, I just followed that.

We should be using double quotes only when string interpolation is needed. I'll see how I could tweak Rubocop to warn about that. Thanks!

PR #91, opened 2016-09-18
Orchestrators

### LINKED ISSUE #65: The Need for Super-Organizers
As our business requirements and code complexity grows, we see the need of weaving together Organizers just like we do with Actions. We have conventions: an Action can not call other Action(s), an Organizer can not invoke other Organizer(s). This is how we try to keep our code simple and easy to follow.

We have been calling multiple Organizers from controller actions which is a nice way of pushing out logic from Rails, but it seems it's not enough. We could build new organizers that reuse all the actions from the other organizers, but I am not sure that's the best solution, as the Organizer is responsible for one specific logic. For example, one Organizer handles Payment, the other communicates to an external API if the Payment was successful.

We've been considering "Orchestrators" or "Super Organizers", that would treat Organizers as higher level Actions. With this change the Organizers have a uniform interface, they might express `expects` and `promises` as well.

Do you have any thoughts on this?

adomokos:
Closes #65. You can find a pretty good discussion there on why this feature is being introduced. Look at README to find out more about this functionality.


README.md
```diff
@@ -571,7 +572,91 @@ end
...
+   |-> run - steps
+   Organizers
+       |-> call - actions
+       Actions
+           |-> execute
+```
+
+You can mix organizers with actions in the orchestrator steps, but mixing other organizers with actions in an organizer is discouraged.
+
+The 4 different constructs an organizer can have are the followings:
```
adomokos: Good catch, thank you!

lib/light-service/orchestrator.rb
```diff
@@ -0,0 +1,79 @@
...
+        lambda do |ctx|
+          ctx = code_block.call(ctx)
+          ctx
+        end
+      end
+
+      def iterate(collection_key, steps)
+        lambda do |ctx|
+          collection = ctx[collection_key]
+          item_key = collection_key.to_s.singularize.to_sym
```
Jared-Prime: I'm trying this branch out in a standalone project, and I noticed that the `String#singularize` comes from `ActiveSupport`. Is that gem a dependency for this feature, or is there another way to singularize the `collection_key`?

adomokos: No, unfortunately. ActiveSupport was added when we introduced I18n message lookups for error or success messages.  I spiked out the removal of AS, and it turned out to be too big of an effort for now. But eventually, I'd like to see no Gem dependency in the project.

kofronpi:
great work, can't wait to use this in production :+1:

adomokos:
Sweet! Thanks for the feedback @kofronpi! Before the year is out, this will make its way to master.

gee-forr:
Hi @adomokos - I see that branch 0.7.0 was merged into this branch... should I use this branch until the next version of LS is released?

adomokos:
Yes, this is the latest release candidate.

adomokos:
Release 0.7.0 went out, this will be in 0.8.0. I'll prepare that branch shortly.

adomokos:
I moved all this code into the [0.8.0](https://github.com/adomokos/light-service/pull/91) PR. It will be released under that version.

I am closing this PR, but won't delete the branch, as developers (myself included) are using the `ad-orchestrators` branch. I'll delete this branch when 0.8.0 is released.

PR #94, opened 2016-10-11
DRY up expects and promises key initializeation logic

adomokos:
I noticed we have the `expected_keys` and `promised_keys` initialization
logic twice, removing the duplication with this commit.


padi:
:+1:

PR #95, opened 2016-10-17
Return a false value from a context when using [] syntax.

bschmeck:


spec/context_spec.rb
```diff
@@ -157,14 +157,20 @@
...
+  it "can contain false values" do
+    context = LightService::Context.make(:foo => false)
+    expect(context[:foo]).to eq false
+  end
+
   context "when aliases are included via .make" do
     let(:context) do
       LightService::Context.make(
         :foo => "foobar",
+        :foo2 => false,
```
adomokos: I think you meant `:bar2` here. When I run the specs as it is, it passes with the old code, but when I change this key to `:bar2`, it fails with receiving nil instead of `false`.

bschmeck: I believe it should be `:foo2`.  The test ensures that you can retrieve a false value via the alias, which has always worked.  (Looking up a false value via the key failed because the second half of the `||` would execute and overwrite the retrieved value.)

I can remove the test if you don't think it's helpful.

bschmeck: The test fails if you change it to `:bar2` and without changing the alias, because (I believe) all keys that match aliases get set to nil.  That behavior is causing a spec failure elsewhere, I'll try to  take a look at it later tonight.

adomokos:
This change will make the tests pass:
```ruby
def [](key)
  hash_value = super(key) if key?(key)
  return hash_value unless hash_value.nil?

  super(aliases.key(key))
end
```
I don't think this is the best solution, as it only makes assumptions on the values. What if the hash_value is set to nil?

bschmeck:
Pushed a change to use the key's alias if there is one, otherwise just do a normal lookup.

Because aliases are keys in the context in their own right, this method could be removed and the standard `Hash#[]` method could be used.  But that causes a failing spec due to an inconsistency with aliases.  Working on that in a separate branch.

bschmeck:
Branch showing issues with context aliases is here: https://github.com/bschmeck/light-service/tree/context-alias-bugs

Diff is here: https://github.com/adomokos/light-service/compare/master...bschmeck:context-alias-bugs

adomokos:
Thank you! I won't release a new version until you (we) have a chance to resolve that alias bug. 👍

PR #108, opened 2017-04-25
0.8.1

adomokos:
* Renaming `skip_all!` to `skip_remaining!`
This warning is issued:
```shell
DEPRECATION WARNING: Using skip_all! has been deprecated, please use `skip_remaining!`
instead. (called from skip_all! at /home/ubuntu/importer/vendor/cache/light-service-
dc0c602cb229/lib/light-service/context.rb:98)
```
* Introducing `ContextFactory` for easier test data setup


lib/light-service/context.rb
```diff
@@ -45,13 +46,13 @@ def failure?
       success? == false
     end
 
-    def skip_all?
-      @skip_all
+    def skip_remaining?
```
padi: 👍

PR #110, opened 2017-06-13
Inspect the context

adomokos:
I noticed that the output in Pry is a bit "minimalistic".

pry_output

Pry uses `pp` to print the object, and since the Context is a glorified hash, it prints it as a hash.

This PR customizes the `inspect` method on the `LightService::Context` object.

This is  what we had before:

```shell
[1] pry(#<RSpec::ExampleGroups::LightServiceContext::CanBeMade::WithFAILURE>)>
context
=> {}
[2] pry(#<RSpec::ExampleGroups::LightServiceContext::CanBeMade::WithFAILURE>)> 
context.inspect
=> "{}"
```

And this is after:

```shell
[1] pry(#<RSpec::ExampleGroups::LightServiceContext::CanBeMade::WithFAILURE>)> context
=> {}
[2] pry(#<RSpec::ExampleGroups::LightServiceContext::CanBeMade::WithFAILURE>)>
context.inspect
=> "LightService::Context({}, success: false, message: '', error_code: nil, skip_remaining: false, aliases: nil)"
```

It would be nice to get a better print output for pry's pp call, but that will be a new PR.


padi:
👍

PR #113, opened 2017-07-20
Short circuit orchestrator methods

timferro:
When context is failed or set to skip remaining.


adomokos:
I tested this PR against 2 complex projects that use LS, passed all our specs. 👍

adomokos:
Thanks for it!

Issue #62, opened 2015-05-18
Rename LightService from light-service to light_service


adomokos:
This is the recommended naming pattern. I chose `light-service` based on a blog post I read at the time when I started LightService, but I think now that the name is misleading.

I am not alone with this unique naming convention. The project [RestClient](https://github.com/rest-client/rest-client) follows the same naming convention.

I would be curious to see what other folks are thinking about the possibility of renaming the gem from `light-service` to `light_service`.

padi:
Also noticed this before. Never bothered opening an issue for it, but :+1:

ramontayag:
Yes, `light_service`, since the project isn't `Light::Service`.

variousred:
:+1:

adomokos:
This is actually not all that easy. There is no rename in RubyGems, I have to close the current LightService repo, open up a new one with the new name and continue development on that. Which might make sense as we approach 1.0.

Does anybody else have a simpler, less invasive resolution to this?

jeroenvisser101:
How about making a new one and publishing to both, abandoning (possibly with deprecation warnings to STDOUT) the old?

adomokos:
That's a pretty good idea! Thank you for suggesting that.

jeroenvisser101:
@adomokos thanks — I've not tested that yet, but in theory, it should work.

gee-forr:
Apologies for the bump on an ancient issue, but I would also add a `spec.post_install_message` to the gemspec for the deprecated version.

There are [docs for post_install_message](http://rubygems.rubyforge.org/rubygems-update/Gem/Specification.html) functionality.

adomokos:
I think we know what we want to do with this. I am proposing closing it for now.

adomokos:
I am not sure if we should proceed with this. Closing it, but please reopen if you think this is a must-have.

PR #116, opened 2017-08-11
Adding `with_callback` to Orchestrators

adomokos:
See the README for more detail.

### DIFF
--- a/lib/light-service/orchestrator.rb
+++ b/lib/light-service/orchestrator.rb
@@ -70,7 +70,9 @@ module LightService
 
       def with_callback(action, steps)
         lambda do |ctx|
-          ctx.reset_skip_remaining! unless ctx.failure?
+          return ctx if ctx.stop_processing?
+
+          # This will only allow 2 level deep nesting of callbacks
           previous_callback = ctx[:callback]
 
           ctx[:callback] = lambda do |context|
--- a/lib/light-service/orchestrator.rb
+++ b/lib/light-service/orchestrator.rb
@@ -68,6 +68,22 @@ module LightService
         end
       end
 
+      def with_callback(action, steps)
+        lambda do |ctx|
+          ctx.reset_skip_remaining! unless ctx.failure?
+          previous_callback = ctx[:callback]
+
+          ctx[:callback] = lambda do |context|
+            reduce(steps, context)
+          end
+
+          ctx = action.execute(ctx)
+          ctx[:callback] = previous_callback
+
+          ctx
+        end
+      end
+
       private
 
       def scoped_reduction(ctx, steps)
--- a/spec/acceptance/orchestrator/with_callback_spec.rb
+++ b/spec/acceptance/orchestrator/with_callback_spec.rb
@@ -82,10 +82,10 @@ describe LightService::Orchestrator do
         [
           SetUpNestedContextAction,
           with_callback(IterateOuterCollectionAction,
-                        [ IncrementOuterCountAction,
-                          with_callback(IterateCollectionAction,
-                                        [IncrementCountAction,
-                                         AddToTotalAction])])
+                        [IncrementOuterCountAction,
+                         with_callback(IterateCollectionAction,
+                                       [IncrementCountAction,
+                                        AddToTotalAction])])
         ]
       end
     end
--- a/spec/acceptance/orchestrator/with_callback_spec.rb
+++ b/spec/acceptance/orchestrator/with_callback_spec.rb
@@ -70,7 +70,7 @@ describe LightService::Orchestrator do
     end
   end
 
-  describe 'a more complex example with nexted callbacks' do
+  describe 'a more complex example with nested callbacks' do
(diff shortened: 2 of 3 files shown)


spec/acceptance/orchestrator/with_callback_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+    extend LightService::Orchestrator
+
+    def self.run(context = {})
+      with(context).reduce(steps)
+    end
+
+    def self.steps
+      [
+        SetUpContextAction,
+        with_callback(IterateCollectionAction,
```
padi: Hi! It's been a while since I made some feedback to this project. We're glad to use light-service again in a completely new project that does some ETL work as well. 🥂 🎆 

With that said, It took me a good 10-20 minutes to re-read the spec and source code and I still don't understand it. It helps that you're not here to explain it as to test whether it's readable on its own or not.

I thought that the callback was `InteractionColllectionAction` and not `[IncrementCountAction, AddToTotalAction]` (or did I get this incorrectly?). So many things are at play here: 

1. Maybe callbacks should have their own naming convention akin to naming actions as `*Action`, since the example might be too simple. Or at the very least, make it explicit which classes are the callbacks.

2. Reading `def with_callback(action, steps)` alone in `orchestrator.rb`, it seems to suggest, by proximity, that `action` is the callback, but in this example, `IterationCollectionAction` seems to be doing the main thing (to iterate over numbers) and the counters seems to be doing the side tasks (updating counters). Also `IterationCollectionAction` ... `expects :callback`, which tells me that this is not the callback itself.

Maybe if the callback API is something like this:

```ruby
reduce(
  IterateCollectionAction,
  callback: [IncrementCountAction, AddToTotalAction]
)
```
... it would make it more explicit which are the callbacks.

lib/light-service/orchestrator.rb
```diff
@@ -68,6 +68,22 @@ def iterate(collection_key, steps)
         end
       end
 
+      def with_callback(action, steps)
+        lambda do |ctx|
+          ctx.reset_skip_remaining! unless ctx.failure?
+          previous_callback = ctx[:callback]
+
+          ctx[:callback] = lambda do |context|
```
padi: Update: This is the _critical line of code_ that made me sure of myself that `steps` are the callbacks. With that said, it can be so easy to gloss over once you're used to `steps` being the main `steps`

adomokos:
@padi - thanks for your feedback.

I understand that the code is a bit hard to crack, however, the `with_callback(controll_logic, steps)` signature follows the convention we have for `iterate`, `reduce_until`, `reduce_if`: the first argument is where the control logic lives, and the second one is the actions or steps that will be invoked when the control logic directs it. I am not sure I'd change that.

I am looking at ways to simplify the logic in the `with_callback` method within the [orchestrator](https://github.com/adomokos/light-service/blob/3da2f9b259cf8f837618df5afb0ab351803e9e17/lib/light-service/orchestrator.rb) like you suggested. It's unclear how the `callback` is being assigned.

Also, feel free to submit a PR against this PR, I'd like to use your ideas if they are better than what you see here.

This functionality is very helpful for us to parse 2-5 GB files and process them with a Sax parser. The benefit of not keeping large data in memory is enormous.

adomokos:
@padi - I am going to merge this as is. Would you be against it? Do you have better ideas on how to make this simpler for the developer to understand?

spec/acceptance/orchestrator/with_callback_spec.rb
```diff
@@ -0,0 +1,169 @@
...
+  describe 'a simple case with a single callback' do
+    it 'calls the actions defined with callback' do
+      result = TestWithCallback.run
+
+      expect(result.counter).to eq(3)
+      expect(result.total).to eq(6)
+    end
+  end
+
+  describe 'a more complex example with nexted callbacks' do
```
padi: nexted -> nested?

adomokos: Good catch! Fixing it.

padi:
Nah, I'm okay with it for now. Reading the source is still probably ok for now.

I haven't used maybe half of the features yet anyway. We've gone far with Organizers, Actions and Organizer-Actions in the last project. We're trying out Organizers only recently in a new one that does some ETL work as well.

I'm actually thinking if we can split the light-service project into different modules and add them whenever we need them. I just need to think through whichs ones can live as extensions of LightService and whichs ones are part of the core light-service. Might be a good chance to rename light_service as well #62

adomokos:
Thanks for the feedback.

As we use the gem (and as more developer uses it), I realized that the separation of `Orchestrators` and `Organizers` is often misunderstood and confusing. I'd like to copy the Orchestrator features to Organizers and just kill the Orchestrator concept as it is. You don't have to use `reduce_if` in your Organizers, but if you want to, they are there.

I think that will be done in a new repo, addressing #62, with version 1.0.

PR #118, opened 2017-09-18
Raise an error if expected keys are unused

bschmeck:
This change will cause LightService to raise an error if an action doesn't use one or more of the `expects` keys.  If there are keys that are possibly unused (e.g. as a result of a conditional,) they can be marked as `maybe` using a new macro.

The motivation for adding this is to help ensure that `expects` blocks are accurate, which will help code discovery, grepping, etc.

Possibly this should be optional/configurable behavior?


spec/test_doubles.rb
```diff
@@ -123,6 +123,31 @@ class MakesTeaWithMilkAction
     end
   end
 
+  class MakesTeaMaybeWithMilkAction
+    extend LightService::Action
+    expects :tea, :milk, :use_milk
+    promises :milk_tea
+    maybe :milk
```
adomokos: This observation is about the DSL:

My first intuition was that the `maybe` macro is at the same level as where `expects` and `promises` are. I realized they are not when I removed `:milk` from the expected keys. I have to expect a key that might be listed in `:maybe`.

It seems the `:maybe` is coupled to the `:expects` macro, I wonder if it could be expressed as such.
What do you think about this:
```ruby
class MakesTeaMaybeWithMilkAction
  extend LightService::Action
  expects :tea, {:maybe => :milk}, :use_milk
  promises :milk_tea
  ...
end
```

Or if you want the distinction, I should be able to list keys that only `:maybe` used and not fully expect it.

What do you think?

adomokos: Also, would `:maybe` work on `:promised` keys as well?

bschmeck: I agree that having to list the keys in two different places isn't the best approach here.  I think that what makes it tricky is that this PR is introducing a new concept.  Originally, `expects` just meant "This key needs to be in the context when this action is executed."  But with this change, it now means "This key needs to be in the context and must be accessed when the action is executed" and `maybe` means "This key needs to be in the context and might be accessed when the action is executed."  So we need to find a way to cleanly combine those concepts.

There's also a question of, if a key is expected but not used, does it have to exist in the context?  I think that the answer is yes, we want to be as explicit as possible about the inputs and outputs of an action.  In that light, the term `maybe` might cause confusion, because it could imply that a key doesn't have to be present in the context.  I don't have a good alternative, though.  I've thought of `conditional`, `sometimes`, `ignore` and `maybe_used`.  (`Maybe` also has an entirely different meaning in Haskell, which lead to confusion as well.)

A couple of possibilities for improving the syntax (I'll keep using `maybe` here for consistency):
* Completely separate `expects` and `maybe`, each key is declared in one or the other, but not both.  We would verify that both types of keys exist at the start of execution but would only verify that `expects` keys were accessed during execution.  The above example would become:
```
class MakesTeaMaybeWithMilkAction
  extend LightService::Action
  expects :tea, :use_milk
  maybe :milk
  promises :milk_tea
  ...
end
```

* Let keys be declared as hashes, as you suggested.  I think it might be cleaner to swap your suggested key and value, though:
```
class MakesTeaMaybeWithMilkAction
  extend LightService::Action
  expects :tea, {:milk => :maybe}, :use_milk
  promises :milk_tea
end
```

* Add `:maybe` as a keyword arg to the end of the `expects` declaration.  I think that I like this syntax the best:
```
class MakesTeaMaybeWithMilkAction
  extend LightService::Action
  expects :tea, :use_milk, :maybe => [:milk]
  promises :milk_tea
end
```

* A fourth option would be to remove `maybe` altogether and force all actions to access their `expected` keys.  That would encourage fewer conditionals in actions, but might be too onerous a restriction to place on users.  
I don't think this should work with `promises`, because I like the idea that executing an action will always add specific keys to the context, and LS will raise an error if that contract is violated.  Allowing actions to only sometimes add keys to the context would complicate any actions that are executed later, because now those actions have to code defensively in case the upstream action didn't add a key to the context.

adomokos:
Thank you for the detailed response!

I do like the concept of `:maybe`, this PR will be merged.
I also see the value of providing a higher level check of `:expects`: it's not enough for it to be in the context as a key, the user has to interact with it in the action.

I am not a fan of placing `:maybe` on the same level with `:expects` and `:promises`as it adds another layer of complexity that will be hard to grok for the user.

I like your 3rd suggestion the best as well, as it signals the keys you defined there are special type of `:expects`. I wonder if we could write this for a single item `:maybe`:

```ruby
class MakesTeaMaybeWithMilkAction
  extend LightService::Action
  expects :tea, :use_milk, :maybe => :milk
  promises :milk_tea
end
```

And this if there are multiple `:maybe` items:
```ruby
class MakesTeaMaybeWithMilkAction
  extend LightService::Action
  expects :tea, :use_milk, :maybe => [:milk, :honey]
  promises :milk_tea
end
```

We will need to add detailed examples in the specs where it's obvious how the `expects` is changing. I am happy to help with that as well. 😄

bschmeck:
I've updated the code so that specifying `maybe` keys happens as a kwarg to the `expects` declaration.  The kwarg can be a single key `maybe: milk` or an array of keys `maybe: [:milk, :tea]`.

The implementation doesn't prevent you from re-declaring a key.  E.g. this is legal: 
```
expects :milk, :maybe => :milk
```
and results in the key being considered a `maybe`.

adomokos:
This looks great, thank you for your work on it!

Since this change will have breaking changes, I'll merge it into the 1.0rc1 branch. I'd also like to eliminate orchestrators and move the `reduce_if` etc. macros to organizers. That will be a nice 1.0 release. I'll add an issue for that soon.

adomokos:
I like to start my development on LS (or any other app) with a failing feature spec, that would describe the desired behavior. I did not see new feature specs for this change.

I'd document:
* What happens when you use only maybe keys
* When you use expected keys with maybes
* Only expected keys and no maybes.

I usually point to those specs from the documentation.
I am happy to add those myself.

bschmeck:
Sorry for the delay.  I finally had a chance to add some tests in `spec/acceptance`.  Let me know if those are close to what you're looking for.

adomokos:
Thank you!!

PR #119, opened 2017-10-23
Refactor KeyVerifiers

bschmeck:
This PR simplifies the verifiers used to check that expected keys exist, that expected keys are used and that promised keys are created.  It does a few things:

* Verifiers are expected to return a list of `offending_keys`.  If there are keys in the list, an error will be raised.
* A default formatting for `offending_keys` is provided, as is a default error message.
* An instance method `throw_error?` is used instead of `throw_error_predicate`.
* Duplicated code has been moved to the superclass, but a Verifier is free to override any necessary methods.

### DIFF
--- a/lib/light-service/context/key_verifier.rb
+++ b/lib/light-service/context/key_verifier.rb
@@ -8,13 +8,7 @@ module LightService
         @action = action
       end
 
-      def are_all_keys_in_context?(keys)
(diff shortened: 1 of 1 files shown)


adomokos:
Very nice refactor! 💯  👍 
Thank you for it, Ben!

PR #120, opened 2017-11-10
Make logging more efficient if it isn't turned on

toofishes:
The ruby logger class has two forms- one with method params, and one with a
block. They are different as the block form is lazy evaulated, and only
evaulated if the log level is actually enabled.

We can use this block form logger, as well as the `logger.info?` guard, to save
ourselves from building several log messages if we aren't actually going to
send or print them somewhere. This should have no effect whatsoever on the
contents of logs, it simply saves the overhead of string operations if we
aren't actually logging.

http://brian.pontarelli.com/2006/07/03/ruby-logging-and-performance/

### DIFF
--- a/lib/light-service/configuration.rb
+++ b/lib/light-service/configuration.rb
@@ -16,7 +16,7 @@ module LightService
 
       def _default_logger
         logger = Logger.new("/dev/null")
-        logger.level = Logger::INFO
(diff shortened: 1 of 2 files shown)


adomokos:
Real neat! Thank you for it!

PR #123, opened 2017-12-12
Only pass default argument to Hash#fetch in context if no block given

robwilliams:
Hash#fetch outputs a warning if the arg count is 2 and a block is given.

It's not possible to just pass a block without this change as the `key` and `default_or_block` arguments are always passed to the parent method.

Checking if there is a block and only passing the default if there is no block stops the warning message from showing.


adomokos:
Thank you!! 👍

Issue #127, opened 2017-12-28
after_fail callback


rudyrigot:
To make my test failures more meaningful when asserting `result.success?`, I'd like to throw an error when `ctx.fail!` is called in my test environment. I believe this is best done by adding a callback/hook to an orchestrator that would run when the orchestrator fails. I'm opening this issue to present an idea of how that could work, and welcome people to discuss it (and then I'd be glad to implement it).

Would we rather have something naive like this:

```
class CalculatesTax
  extend LightService::Organizer

  def self.call(order)
    with(:order => order).reduce(
        LooksUpTaxPercentageAction,
        CalculatesOrderTaxAction,
        ProvidesFreeShippingAction
      )
  end

  def self.after_fail
    # Do something
  end
end
```

or something a bit more like Active Record does callbacks, like this:

```
class CalculatesTax
  extend LightService::Organizer

  after_fail :do_something

  def self.call(order)
    with(:order => order).reduce(
        LooksUpTaxPercentageAction,
        CalculatesOrderTaxAction,
        ProvidesFreeShippingAction
      )
  end

  def self.do_something
    # Do something
  end
end
```

I'm not sure a given orchestrator is likely to have as many reasons to stack several same-type callbacks on top of each other, so I'm not sure anything beyond the naive solution is needed; but I'd welcome people's take about it.

rudyrigot:
Closing this because #133 attempts to solve it, conversation can live there.

PR #133, opened 2018-01-05
After_failing callback for orchestrators

rudyrigot:
To make my test failures more meaningful when asserting `result.success?`, I'd like to throw an error when `ctx.fail!` is called in my test environment. I believe this is best done by adding a callback/hook to an orchestrator that would run when the orchestrator fails.

This PR offers a possible implementation for it, but is open to discussion. It is significantly different from the proposals I had made in #127, because after better understanding how LightService is used, I feel it makes more sense to declare it in the line of code that defines what the orchestrator does, rather than the way Active Record does callbacks.

So, with this PR, it works like that:

```
class CalculatesTax

  def self.call(order)
    with(:order => order).after_failing(:foo, :bar).reduce(
        LooksUpTaxPercentageAction,
        CalculatesOrderTaxAction,
        ProvidesFreeShippingAction
      )
  end

  def self.foo(ctx)
    # Do something
  end

  def self.bar(ctx)
    # Do something
  end
end
```

### DIFF
--- a/lib/light-service/context.rb
+++ b/lib/light-service/context.rb
@@ -8,7 +8,8 @@ module LightService
 
   # rubocop:disable ClassLength
   class Context < Hash
-    attr_accessor :message, :error_code, :current_action, :orchestrator_callbacks
+    attr_accessor :message, :error_code, :current_action,
+                  :orchestrator_callbacks
 
     def initialize(context = {},
                    outcome = Outcomes::SUCCESS,
@@ -86,11 +87,8 @@ module LightService
                                                             options)
       @error_code = error_code
       @outcome = Outcomes::FAILURE
-      unless @orchestrator_callbacks[:after_failing].nil?
-        @orchestrator_callbacks[:after_failing].each do |cb|
-          cb.call(self)
-        end
-      end
+
+      run_callbacks :after_failing
     end
 
     def fail_and_return!(*args)
@@ -169,6 +167,13 @@ module LightService
     def check_nil(value)
       return 'nil' unless value
       "'#{value}'"
+    end
+
+    def run_callbacks(symbol)
+      return if @orchestrator_callbacks[symbol].nil?
+      @orchestrator_callbacks[symbol].each do |cb|
+        cb.call(self)
+      end
     end
   end
   # rubocop:enable ClassLength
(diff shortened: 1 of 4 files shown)


lib/light-service/context.rb
```diff
@@ -8,16 +8,18 @@ module Outcomes
...
   class Context < Hash
-    attr_accessor :message, :error_code, :current_action
+    attr_accessor :message, :error_code, :current_action, :orchestrator_callbacks
 
     def initialize(context = {},
                    outcome = Outcomes::SUCCESS,
                    message = '',
-                   error_code = nil)
+                   error_code = nil,
+                   orchestrator_callbacks = {})
```
rudyrigot: Storing the lambdas in `context.orchestrator_callbacks[:after_failing]` to leave the door open to other kinds of callbacks later.
Also, `orchestrator_callbacks` and not `callbacks` in order to disambiguate with the `with_callback` feature.

lib/light-service/orchestrator.rb
```diff
@@ -86,6 +86,15 @@ def with_callback(action, steps)
         end
       end
 
+      def after_failing(symbols)
+        symbols = [symbols] unless symbols.is_a?(Array)
+        @context.orchestrator_callbacks[:after_failing] =
+          symbols.map do |symbol|
+            method(symbol)
```
rudyrigot: The `method` method turns a class method into a lambda. Seriously, awesome Ruby......

adomokos:
@rudyrigot - thank you for this PR!

I am not sure what the real benefit of adding this hook to LS besides supporting tests. I am hesitant to introduce another method to Organizers that could be achieved with (I assume) [around](https://github.com/adomokos/light-service#benchmarking-actions-with-around-advice) or Action `before` and `after` hooks. These do not exist yet, but I'd like to expose them in the near future.

Also, would you add methods to your Orchestrator to support testing?

Let's see how this might be used in our project(s), and if the reason is valid, I'll merge it.

rudyrigot:
No problem at all, I’m not convinced it’s a good idea either after all (or more to the point, that the complexity of this solution would be worth what it brings and what it solves for us). No regret though if we change course, it was great diving into the LightService codebase, and I had fun doing it!
I’m pausing this work for now, let’s figure out what we want to be doing first.

adomokos:
Based on our conversation, I am closing this. Thanks for submitting it, though!

Issue #138, opened 2018-03-15
ContextFactory fully executes an Organizer when used with callback action


jeremy-hanna:
**Summary:**

Trying to use a `ContextFactory` for testing an Organizer with a `with_callback` (and possibly other methods) will lead to full execution of the Organizer to build the returned context.

**Failure recreation:**

```
# create a new organizer Callback and Action
module TestDoubles
  class CallbackOrganizer
    extend LightService::Organizer

    def self.call(number)
      with(:number => number).reduce(actions)
    end

    def self.actions
      [
        AddsOneAction,
        with_callback(CallbackAction, [
          AddsTwoAction,
          AddsThreeAction
        ])
      ]
    end
  end

  class CallbackAction
    extend LightService::Action
    expects :number, :callback

    executed do |context|
      context.number += 10
      context.number =
        context.callback.call(context).fetch(:number)
    end
  end
end

RSpec.describe TestDoubles::AddsTwoAction do
  it "does not execute a callback entirely" do
    context = LightService::Testing::ContextFactory
                      .make_from(TestDoubles::CallbackOrganizer)
                      .for(described_class)
                      .with(:number => 0)

    # add 1, add 10, then stop before executing first add 2
    expect(context.number).to eq(11)
  end
end

# => Failures:

  1) TestDoubles::AddsTwoAction does not execute a callback entirely from a ContextFactory
     Failure/Error: expect(context.number).to eq(11)

       expected: 11
            got: 16
```

**Technicals:**

What is going on is [the ContextFactory](https://github.com/adomokos/light-service/blob/master/lib/light-service/testing/context_factory.rb#L35) enumerates through the duplicated `organizer.actions`, however when used with a `with_callback(FooAction, [ SomeStepAction ])` the callback action and subsequent steps get flattened into a lazily executed Proc [here](https://github.com/adomokos/light-service/blob/master/lib/light-service/organizer.rb#L46).

Therefore a ContextFactoryOrganizer.organizer `.actions` will never have the action constant in the array to match on.

**Proposed Solution:**
The ContextFactoryOrganizer is going to have to get smarter and build out an Action execution path tree for these steps and wrap them in a new organizer that halts at the expected Action

adomokos:
Thank you for filing this issue. When I pushed logic from Orchestrators to Organizers, I don't think I tested all the different components, like `ContextFactory`. It works with the original Organizer methods, but might not work with the new ones.

I'll look into it or PRs to fix issues are happily accepted!

[...]

PR #141, opened 2018-03-20
ContextFactory patch for Organizer methods

jeremy-hanna:
## Summary

As detailed in https://github.com/adomokos/light-service/issues/138 - Organizer methods `with_callback`, `iterate`, `reduce_if`, and `reduce_until` break a ContextFactory context creation because at runtime these get bundled into a lazy Proc to be run, loosing the ability to pick and choose which `actions` to execute and effectively making the ContextFactory execute the entire Organizer to generate a context.

There weren't specs, so I added a bunch and broke them off into directories.

~~The proposed solution is not all that pretty as I wasn't sure if I could get it even to work - well it works and I'm unsure how you'd like to style it. The approach consists of:~~

~~1. lexically parse the _source_ `actions` method of an Organizer~~
~~2. rebuilding the correct namespace of actions~~
~~3. grouping the subset of actions each method uses as steps~~
~~4. re-evaluating the runtime Proc of these methods~~

~~Since all of this is handily done at Ruby syntax evaluation, we effectively need to _redo_ that step again _in test runtime_.~~

~~The collection of actions are really just an ordered list of execution actions, instead of restructuring this for the current implementation into a execution or parse tree for actions - there is just a lot of array index manipulation (which is not ideal, but structurally I wanted to keep these as similar as possible to an actual Organizer).~~

~~OF NOTE: it is fundamentally impossible to recreate a `iterate` or a `reduce_until` method in this way if the action we are building a context _up to_ is within that enumeration as this _has to_ close the loop.~~

Using a `before_action` lambda introduced in https://github.com/adomokos/light-service/pull/144 , we can catch and throw an escape from execution when the Organizer gets to the expected action.


adomokos:
@jeremy-hanna: thank you so much for putting a LOT of work into this PR!! 💯 

I really like that the updated `ContextFactory` is now working with Orchestrator methods. It was definitely something I did not consider when I ported over functionality from Orchestrators into Organizers. Sorry about that.

I looked at the code and it looks a bit complex considering what it does. In my mind, from the outside, the `ContextFactory` is this black box that takes an Organizer, an initial context and an action class name. It should generate the context up-to that action so you don't have to fiddle with setting up contexts for your action tests.

Your proposed solution is doing heavy parsing of the Organizer, using a lexer, etc. I think it's a brilliant solution, but I wonder if there was a simpler solution.

What if we could hook into the execution pipeline, feed the Organizer the initial arguments or context, let it go, but stop right before the action is being executed. I need to think about how this could be done, but this solution is close to the "Let LightService provide before/after hooks for actions" idea I've been thinking about.

If we had that functionality, the ContextFactory would have to set up the Organizer with the initial context, add a callback hook to the action's `before_execute` method, stop the process right there and just return the context from that point. This way you let the Organizer execute the call chain and no heavy parsing is needed.

What do you think about that idea? If you think this proposed solution with before/after hooks would help, I'd be happy to spike it out.

spec/testing/context_factory_spec.rb
```diff
@@ -30,11 +32,27 @@
...
+        .with(1)
 
       expect(ctx.number).to eq(4)
     end
   end
+
+  context 'when called with an invalid action' do
+    let(:invalid_action) { TestDoubles::MakesLatteAction }
+
+    xit 'raises an argument error' do
```
jeremy-hanna: There's not a way to do this currently with the ContextFactory using `before_action`s

adomokos:
I'll add my before/after hooks PR and once that's in, we can finalize this and merge it.

Thank you @jeremy-hanna for working on it!

lib/light-service/testing/context_factory.rb
```diff
@@ -1,43 +1,35 @@
...
-      def find_up_to(action)
-        original_actions = organizer.actions
-
-        original_actions.take_while do |current_action|
-          current_action != action
-        end
-      end
     end
   end
 end
```
adomokos: I really like how much simpler this class became! 💥

spec/testing/context_factory_spec.rb
adomokos: By introducing a bit more state in the `ContextFactory` we could make this pass:

```ruby
module LightService
  module Testing
    class ContextFactory
      attr_reader :organizer

      def self.make_from(organizer)
        new(organizer)
      end

      def for(action)
        @action_matched = false

        @organizer.before_actions = [
          lambda do |ctx|
            if ctx.current_action == action
              @action_matched = true
              throw(:return_ctx_from_execution, ctx)
            end
          end
        ]

        self
      end

      def with(ctx)
        escaped = catch(:return_ctx_from_execution) do
          @organizer.call(ctx)
        end

        if @action_matched == false
          raise ArgumentError.new "Could not find action in #{self.organizer}"
        end

        escaped
      end

      def initialize(organizer)
        @organizer = organizer
      end
    end
  end
end
```

However, thanks to your testing, [this spec](https://github.com/jeremy-hanna/light-service/blob/380290964c33e1fec6045351ecd98af56add152c/spec/testing/context_factory/reduce_if_spec.rb#L30-L38) fails.

Is this something we could tweak here?

I don't want to have pending specs in LightService and not having a solution for it.

jeremy-hanna: If you want to add state to this class for this functionality, please feel free to do so as an append to this PR or in another - I was only mentioning something that my solution would not achieve and possibly indicate a passing context to the user (developer) when in reality the organizer executed the entire action set which was not expected

adomokos:
Thank you!! 👍

PR #144, opened 2018-03-26
Before/After Action Hooks

adomokos:
As we used LS for more and more complex workflows (yeah, orchestrator functionality was born), we had to add instrumentation or custom logic before/after some or all actions.

I really liked [AOP concepts](https://en.wikipedia.org/wiki/Aspect-oriented_programming) back in the early 2000s, the idea of injecting custom code (almost like a plug-in interface) fascinated me.

This PR introduces the `before_actions` and `after_actions` hooks in LightService Organizers. These custom hooks at the Organizer level and they fire (if provided) before and after each action. They are great for instrumenting your code, yet you don't have to clutter the business logic with instrumentation.

Consider this code:

```ruby
class Organizer

  def call(ctx)
    with(ctx).reduce(actions)
  end

  def actions
    OneAction,
    TwoAction,
    ThreeAction
  end
end

class TwoAction
  extend LightService::Action
  expects :user, :logger

  executed do |ctx|
    # Logging information
    if ctx.user.role == 'admin'
       ctx.logger.info('admin is doing something')
    end

    ctx.user.do_something
  end
end
```

Note how the logging information is blurring the actual business logic.

You could now do this:

```ruby
Organizer.before_actions = 
  lambda do |ctx|
    if ctx.current_action == TwoAction
      return unless ctx.user.role == 'admin'
      ctx.logger.info('admin is doing something')
    end
  end

class TwoAction
  extend LightService::Action
  expects :user, :logger

  executed do |ctx|
    ctx.user.do_something
  end
end
```

Look, how the instrumentation logic has been decoupled from business logic.

The same is true for `after_actions`.

README update will be attached to this PR.

### DIFF
--- a/spec/acceptance/before_actions_spec.rb
+++ b/spec/acceptance/before_actions_spec.rb
@@ -4,11 +4,10 @@ require 'test_doubles'
 RSpec.describe 'Action before_actions' do
   describe 'works with simple organizers - from outside' do
     it 'can be used to inject code block before each action' do
-      TestDoubles::AdditionOrganizer.before_actions = [
+      TestDoubles::AdditionOrganizer.before_actions =
         lambda do |ctx|
           ctx.number -= 2 if ctx.current_action == TestDoubles::AddsThreeAction
         end
-      ]
 
       result = TestDoubles::AdditionOrganizer.call(0)
 
--- a/spec/acceptance/before_actions_spec.rb
+++ b/spec/acceptance/before_actions_spec.rb
@@ -1,7 +1,7 @@
 require 'spec_helper'
 require 'test_doubles'
 
-RSpec.describe 'Action before hooks' do
+RSpec.describe 'Action before_actions' do
   describe 'works with simple organizers - from outside' do
     it 'can be used to inject code block before each action' do
       TestDoubles::AdditionOrganizer.before_actions = [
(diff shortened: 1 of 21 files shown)


spec/acceptance/before_actions_spec.rb
```diff
@@ -0,0 +1,109 @@
...
+    it 'accepts before_actions hook lambdas from organizer' do
+      result = BeforeActions::AdditionOrganizer.call(0)
+
+      expect(result.fetch(:number)).to eq(1)
+    end
+  end
+
+  describe 'works with callbacks' do
+    it 'can interact with actions from the outside' do
+      TestDoubles::TestWithCallback.before_actions = [
```
rudyrigot: I'm not sure why, but I'm having tingles about allowing people to add those callbacks dynamically at calling time on top of the other way, declared in the organizer. I'm thinking of people wondering "There's a thing going on, and I'm not sure what it is", and then going to their actions and organizer, not seeing it, and not getting where it can be coming from. Not sure if that's relevant.

adomokos: We need to be able to modify the organizer and inject logic from the outside. See [this PR](https://github.com/adomokos/light-service/pull/141). The [modifications](https://github.com/jeremy-hanna/light-service/blob/4381ad539c1c79f246990b6c52ef1943005500e7/lib/light-service/testing/context_factory.rb#L11-L17) to ContextFactory triggered this PR.

The great things about this is that it's there, you don't need to use it if you don't want to. And if you do, use it with great care. Sounds a lot like Ruby metaprogramming. 😉

rudyrigot: Sounds like you thought it through, no problem for me!

PR #148, opened 2018-03-31
Remove extra argument from example

ngpestelos:
Passing `action` as a separate parameter causes an argument error because around_each expects only one argument.

See also: https://github.com/adomokos/light-service/blob/a207d17296d9ff995e39ddeb12ed039159a11606/spec/test_doubles.rb#L54


adomokos:
Thank you! 👍

PR #149, opened 2018-04-20
Organizers call each action with the same context

thegene:
Rather than passing the return of each action call to the subsequent action, organizers now call each action with the same context, allowing actions to mutate the same context action.

Am mainly opening this up for discussion as this _feels_ like a pretty fundamentally breaking change.

The issue this solves for is when an organizer is within another organizer as an action the action proceeding this organizer will be called with the return from `call` rather than the context object. This will of course then fail because the action is being called with something other than a `LightService::Context`.

### DIFF
--- a/lib/light-service/organizer/with_reducer.rb
+++ b/lib/light-service/organizer/with_reducer.rb
@@ -25,7 +25,7 @@ module LightService
         raise "No action(s) were provided" if actions.empty?
         actions.flatten!
 
-        actions.reduce(context) do |current_context, action|
+        actions.each_with_object(context) do |action, current_context|
           begin
             result = invoke_action(current_context, action)
           rescue FailWithRollbackError
--- a/lib/light-service/organizer/with_reducer.rb
+++ b/lib/light-service/organizer/with_reducer.rb
@@ -27,9 +27,9 @@ module LightService
 
         actions.each_with_object(context) do |action, current_context|
           begin
-            result = invoke_action(current_context, action)
+            invoke_action(current_context, action)
           rescue FailWithRollbackError
-            result = reduce_rollback(actions)
+            reduce_rollback(actions)
           ensure
             # For logging
             yield(current_context, action) if block_given?
--- a/lib/light-service/organizer/with_reducer.rb
+++ b/lib/light-service/organizer/with_reducer.rb
@@ -34,8 +34,6 @@ module LightService
             # For logging
             yield(current_context, action) if block_given?
           end
-
-          result
         end
       end
 
--- a/spec/test_doubles.rb
+++ b/spec/test_doubles.rb
@@ -102,6 +102,32 @@ module TestDoubles
     end
   end
 
+  class NestedOrganizer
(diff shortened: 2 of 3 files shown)


adomokos:
Interesting change. I've just ran your proposed change on our code base and it passed all the tests. Let's chat about it and figure out if we want this to be merged or not.

Thanks for submitting it!

lib/light-service/organizer/with_reducer.rb
```diff
@@ -25,7 +25,7 @@ def reduce(*actions)
         raise "No action(s) were provided" if actions.empty?
         actions.flatten!
 
-        actions.reduce(context) do |current_context, action|
+        actions.each_with_object(context) do |action, current_context|
```
adomokos: By using `each_with_object`, you can remove the `result` from line 38. That's one of the benefits of using `each_with_object`.
I tested one of our apps with that change, and it passed all the tests.

thegene: good catch thanks

spec/test_doubles.rb
```diff
@@ -102,6 +102,32 @@ def self.do_something_with_no_starting_context
...
+    def self.call(context)
+      context[:foo] = [1, 2, 3]
+    end
+  end
+
+  class NestingOrganizer
+    extend LightService::Organizer
+
+    def self.call(context)
+      with(context).reduce([NestedOrganizer, NestedAction])
```
thegene: This basically reproduces the issue we were seeing

spec/test_doubles.rb
```diff
@@ -102,6 +102,32 @@ def self.do_something_with_no_starting_context
     end
   end
 
+  class NestedOrganizer
+    extend LightService::Organizer
+
+    def self.call(context)
+      context[:foo] = [1, 2, 3]
```
thegene: Here is my main underlying question: is an Organizer supposed to basically also be an action, or is it supposed to apply a bunch of actions on, and then return, a context?

adomokos:
@thegene - I'd like to final review and merge this PR. Can you please make Rubocop happy?

adomokos:
We might need to add a bit better specs to this change.

I grabbed your changes and reset the `WithReducer` class to be the same as we have it in master right now. I switched back to `reduce` from `each_with_object`. After I polished the specs you added, all your new specs were passing with the original code. See my code tweak [here](https://github.com/adomokos/light-service/commit/5ca4bb1592be4134b147b55a458d2d1e79e61516).

The `ReturningOrganizer` just sets the `:foo` value in the context, but it was not returning the context. Now I am realizing, that by not returning the context there is the benefit of using `each_with_object`, which is the whole point of this PR.
👍 💯 
I have never used Organizers without Actions, I guess that's why I was a bit confused by it.

I'd suggest renaming that organizer to be `NotExplicitlyReturningContextOrganizer`, to be more clear. What do you think?

adomokos:
Thank you!!! :100:

PR #156, opened 2018-08-08
Change how some methods are documented.

mrkjlchvz:
## Description
While trying to learn the API, I found that some methods were not written exactly the way 
it should be. Is there something I'm missing here?

## Changes
- Instead of `call`, use `self.call`
- Use `self.actions` instead of `actions`

BTW, I find this library useful and I've used it for a number of projects I'm working on. Keep up the good work!

### DIFF
--- a/README.md
+++ b/README.md
@@ -294,14 +294,16 @@ Consider this code:
 class SomeOrganizer
   extend LightService::Organizer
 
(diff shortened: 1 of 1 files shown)


adomokos:
Thank you so much for this PR! I am glad the documentation is getting better.

PR #157, opened 2018-08-09
Move `require 'active_support/core_ext/string'` to gem files from test files

### LINKED ISSUE #155: undefined method `singularize'
When `active_support/core_ext/string/inflections` is not required, I get the error above when using `iterate`. Perhaps active support should be a dependency, or at least the gem complains that it must be required if `iterate` is used so that the developer can add the gem to the list of dependencies.

Thoughts?

ramontayag:
Fix #155

### DIFF
--- a/lib/light-service.rb
+++ b/lib/light-service.rb
@@ -1,4 +1,5 @@
 require 'logger'
+require 'active_support/core_ext/string'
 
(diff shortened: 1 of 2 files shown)


adomokos:
Thank you!

Issue #159, opened 2018-10-05
Per-organizer logging


ramontayag:
@adomokos, what are you thoughts on per-organizer logging?

I find logging to be helpful, especially on a newly deployed organizer to see what problems may occur. However, since LightService is used extensively, the logs have become too verbose, logging about other organizers/actions that I no longer care to see in detail.

adomokos:
I like it! Care to start a PR with it?

ramontayag:
Yup, I am happy to do it, but I will be able to start in about a week or two. :)

adomokos:
Great! Maybe submit a PR early on,  even with incomplete code, to make sure you are going down the right path. Thanks for it in advance!

adomokos:
I'll release it shortly, I just want to sneak in another deprecation warning for key aliases.

PR #162, opened 2018-11-02
Per organizer logger

ramontayag:
Addresses https://github.com/adomokos/light-service/issues/159

### DIFF
--- a/README.md
+++ b/README.md
@@ -538,6 +538,15 @@ I, [DATE]  INFO -- : [LightService] - ;-) <TestDoubles::MakesLatteAction> has de
 I, [DATE]  INFO -- : [LightService] - context message: Can't make a latte with a fatty milk like that!
 ```
 
+You can specify the logger on the organizer level, so the organizer does not use the global logger.
+
+```ruby
+class FooOrganizer
+  extend LightService::Organizer
+  log_with Logger.new("/my/special.log")
+end
+```
+
 ## Error Codes
 You can add some more structure to your error handling by taking advantage of error codes in the context.
 Normally, when something goes wrong in your actions, you fail the process by setting the context to failure:
--- a/lib/light-service/organizer.rb
+++ b/lib/light-service/organizer.rb
@@ -56,6 +56,14 @@ module LightService
       def with_callback(action, steps)
         WithCallback.run(self, action, steps)
       end
+
+      def log_with(logger)
+        @logger = logger
+      end
+
+      def logger
+        @logger
+      end
     end
 
     module Macros
--- a/lib/light-service/organizer/with_reducer_factory.rb
+++ b/lib/light-service/organizer/with_reducer_factory.rb
@@ -6,13 +6,13 @@ module LightService
                  LightService::Configuration.logger
         decorated = WithReducer.new
 
-        if logger.nil?
-          # :nocov:
-          WithReducer.new
-          # :nocov:
-        else
-          WithReducerLogDecorator.new(monitored_organizer, WithReducer.new)
-        end
+        return decorated if logger.nil?
+
+        WithReducerLogDecorator.new(
+          monitored_organizer,
+          :decorated => decorated,
+          :logger => logger
+        )
       end
     end
   end
--- a/lib/light-service/organizer/with_reducer_factory.rb
+++ b/lib/light-service/organizer/with_reducer_factory.rb
@@ -2,7 +2,11 @@ module LightService
   module Organizer
     class WithReducerFactory
       def self.make(monitored_organizer)
-        if LightService::Configuration.logger.nil?
+        logger = monitored_organizer.logger ||
+                 LightService::Configuration.logger
+        decorated = WithReducer.new
+
+        if logger.nil?
           # :nocov:
           WithReducer.new
           # :nocov:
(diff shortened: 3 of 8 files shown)


adomokos:
Thanks for working on this. I'll take a look at it today.

spec/acceptance/log_from_organizer_spec.rb
```diff
@@ -151,4 +153,38 @@ def collects_log
...
+      end
+    end
+    it "logs in own logger" do
+      global_logger_organizer.call
+      custom_logger_organizer.call(coffee: "Cappucino")
+
+      # Test that overriden.log was written into
+      binding.pry
+    end
+  end
```
adomokos: I like that you added tests. Can you please move it into a new spec file? Something specific to this feature.

spec/spec_helper.rb
```diff
@@ -18,5 +18,16 @@
...
 
 I18n.enforce_available_locales = true
+
+SPEC_DIR = Pathname.new(File.dirname(__FILE__))
+LOG_DIR = SPEC_DIR.join("../tmp/logs")
+
+RSpec.configure do |c|
+  c.before :each do
+    FileUtils.rm_rf LOG_DIR
+    FileUtils.mkdir_p LOG_DIR
```
adomokos: I don't like IO in the tests. Can we use in-memory logger to avoid this?

lib/light-service/organizer/with_reducer_factory.rb
```diff
@@ -2,7 +2,9 @@ module LightService
   module Organizer
     class WithReducerFactory
       def self.make(monitored_organizer)
-        if LightService::Configuration.logger.nil?
+        logger = monitored_organizer.logger || LightService::Configuration.logger
+
+        if logger.nil?
```
adomokos: I was skeptical about this change first, but it might work. 👍

lib/light-service/organizer.rb
```diff
@@ -56,6 +56,14 @@ def execute(code_block)
       def with_callback(action, steps)
         WithCallback.run(self, action, steps)
       end
+
+      def log_with(logger)
+        @logger = logger
+      end
+
+      def logger
+        @logger
```
adomokos: What do you think about moving the conditional into here?

```ruby
def logger
  @logger || LightService::Configuration.logger
end
```
That way  we can use `logger = monitored_organizer.logger` without further conditionals in `lib/light-service/organizer/with_reducer_factory.rb`.

spec/spec_helper.rb
adomokos: Maybe [this](https://stackoverflow.com/questions/1484129/ruby-send-logger-messages-to-a-string-variable) would work.

ramontayag:
Thanks I'll take a look at this later!

lib/light-service/organizer.rb
ramontayag: Yes, there will be less conditionals in `WithReducerFactory` but the object will be a little bit less "functional" -- if the global logger is changed, the way the classes behave will change during runtime. One could argue that this is desirable.

For the way I use it, it's not a problem, so I defer the decision to you: which do you prefer?

adomokos: I like your suggestion, keep it as is. Thanks!

lib/light-service/organizer/with_reducer_log_decorator.rb
```diff
@@ -5,10 +5,10 @@ class WithReducerLogDecorator
 
       alias logged? logged
 
-      def initialize(organizer, decorated = WithReducer.new)
+      def initialize(organizer, decorated: WithReducer.new, logger:)
```
ramontayag: If this class is used outside the gem, then this would be a breaking change. I can make it backwards compatible if you think it's worth it.

spec/acceptance/custom_log_from_organizer_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+
+    it "logs in own logger" do
+      global_logger_organizer.call(1)
+      custom_logger_organizer.call(:coffee => "Cappucino")
+
+      expect(custom_logger_string.string).to include("MakesLatteAction")
+      expect(custom_logger_string.string).to_not include("AddsOneAction")
+    end
+  end
+end
```
adomokos: This test is great! 👍 

Can you add one more test case, where the default logger goes into one sink (StringIO in thise case), and a custom logger goes into another one? I'd like to make sure it's backward compatible

README.md
```diff
@@ -538,6 +538,15 @@ I, [DATE]  INFO -- : [LightService] - ;-) <TestDoubles::MakesLatteAction> has de
...
 
+You can specify the logger on the organizer level, so the organizer does not use the global logger.
+
+```ruby
+class FooOrganizer
+  extend LightService::Organizer
+  log_with Logger.new("/my/special.log")
+end
+```
+
```
adomokos: Thanks for adding this!

spec/acceptance/custom_log_from_organizer_spec.rb
ramontayag: Of course! Done. :+1:

lib/light-service/organizer/with_reducer_log_decorator.rb
adomokos: It is an internal class, I think it's 👍.

adomokos:
Looks good, thank you for your contribution!

PR #164, opened 2018-12-10
Add tests for skipping inside `reduce_if`

ramontayag:
@adomokos what would you say is the correct behaviour here? Should it stop the rest of the actions?


adomokos:
If I recall correctly, we wanted to scope the `skip_remaining!` within the actual code block and this was the intended behavior. I couldn't find a supporting spec to confirm this. :-(

I would suggest an alternate `skip_remaining!`, like `halt!` or something to stop processing immediately without a failure from any level of the orchestrator.

ramontayag:
I like that. Makes sense to have two versions. I can rename this spec so it reflects what you just mentioned.

adomokos:
Let's change this test to cover the current functionality, so it passes. I'd clone this test for the new method: `halt!`.
FTR: `fail!` should stop everything immediately.

ramontayag:
I forgot that rubocap was failing! @adomokos this passes now except for some Travis build issue.

adomokos:
It seems the build is still failing. Can you plz take a look?

ramontayag:
Ok done @adomokos

adomokos:
Thank you! 👍

and-enaruta:
@adomokos What is the reason to scope `skip_remaining!` in `reduce_if`? I found it a little bit misleading especially when the documentation says "...behaves similarly to `fail!`, except this will not push the context into a failure state".

ramontayag:
@and-enaruta this is if you want to skip just the actions in the `reduce_if`, and not the whole thing.

```ruby
[
  RunsOne,
  RunsTwo,
  reduce_if(true, [
    RunsThree,
    RunsFour, # skip_remaining called here
    DoesNotRun
  ])
  RunsFive,
]
```

and-enaruta:
@adomokos @ramontayag Thank you for your quick response. The question is why `skip_remaining!` need to be scoped by `reduce_if`.

Here is the business case that I ran into:

```ruby
[
  RunsOne,
  RunsTwo,
  reduce_if(feature_toggle_on, [
    RunsThree,
    RunsFour, # skip_remaining called here
  ]),
  RunsFive,
]
```

We hide the code behind feature toggles. In the above implementation `RunsFive` will always run. When the feature toggle will be removed the same code will work differently and start skipping `RunsFive`.

For now, as a quick workaround, I did following:

```ruby
[
  RunsOne,
  RunsTwo,
  reduce_if_else(feature_toggle_on, [
    RunsThree,
    RunsFour, # skip_remaining called here
    RunsFive,
  ], [
    RunsFive,
  ]),
]
```

But it is not scalable when you have multiple feature flags for multiple actions.

adomokos:
Oh, I see, this might be a scoping issue.

PR #172, opened 2019-09-21
Add add_to_context and add_aliases organizer methods

ni3t:
Hi @adomokos!

I'm creating this PR to potentially add two new utility methods upstream that I've found very useful in working with LightService.

The background on why I've found these useful is as follows:

1. It is common for organizers to share actions
2. I've found it quite useful to nest organizers when they perform common series of actions by adding `SomeOtherOrganizer.actions` to the list of actions in another Organizer
3. When pulling actions from another organizer directly, the module macro methods aren't executed and the actions pulled in will likely fail due to keys not in the context.

Therefore, these two simple methods allow an Organizer's actions to be effectively pulled in and executed within another Organizer.

So, instead of this pattern:
```
class OrganizerOne
  aliases(foo: :bar)
	
  def self.call
	with(...).reduce(actions
  end

  def self.actions
	[
	  ...
	]
  end
end
```

This pattern can be used instead:

```
class OrganizerOne

  def self.call
	with(...).reduce(actions)
  end

  def self.actions
	[
	  add_aliases(foo: :bar),
	  ...
	]
  end
```

And this allows another organizer to have the alias in its context like so:

```
class OrganizerTwo
  
  ...

  def self.actions
	[
	  OrganizerOne.actions # alias foo: :bar will now be in the context
	]
  end
```

# edit
since I forgot to mention add_to_context

The second method added, `add_to_context`, is simply DSL sugar that explicitly states that you're adding something to the context.

You can achieve this now by using an execute proc, but the `add_to_context` method makes it more explicit what you're doing.

So this:

`execute( ->(ctx) { add_to_context(foo: :bar, baz: :qux) }`

can now be executed directly as an action macro:

`add_to_context(foo: :bar, baz: :qux)`

Thanks again, look forward to hearing from you!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,4 @@ test/version_tmp
 tmp
 vendor/bundle
 bin
+.idea
--- a/lib/light-service/organizer.rb
+++ b/lib/light-service/organizer.rb
@@ -64,6 +64,16 @@ module LightService
       def logger
         @logger
       end
+
+      def add_to_context(**args)
+        args.map do |key, value|
+          execute(->(ctx) { ctx[key.to_sym] = value })
+        end
+      end
+
+      def add_aliases(args)
+        execute(->(ctx) { ctx.assign_aliases(ctx.aliases.merge(args)) })
+      end
     end
 
     module Macros
--- a/spec/organizer_spec.rb
+++ b/spec/organizer_spec.rb
@@ -90,4 +90,18 @@ describe LightService::Organizer do
       expect(reduced).to eq(ctx)
     end
   end
+
+  context "#add_to_context" do
+    it 'should add to the context' do
+      result = TestDoubles::AnOrganizerThatAddsToContext.call
+      expect(result[:strongest_avenger]).to eq :thor
+    end
+  end
+
+  context "#add_aliases" do
+    it 'should add to the aliases' do
+      result = TestDoubles::AnOrganizerThatAddsAliases.call
+      expect(result[:baz]).to eq :bar
+    end
+  end
 end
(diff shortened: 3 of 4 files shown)


adomokos:
@ni3t, thank you for contributing!

You have touched on a very good point. I've [proposed](https://github.com/adomokos/light-service#key-aliases) removing the `aliases` macro a while back, as I thought a lambda could be easily used for that. You are suggesting a more elegant solution which I like. 👍 

Can you give me a more specific example where you'd use `add_to_context`? I don't recall having a need for it, but you might have found a neat solution to something I have not encountered with.

My big question is: would this change allow me to phase out the `aliases` macro. I am asking @gee-forr, since he claimed he used it. I wouldn't remove it right away, I'd introduce deprecation warnings in a new version and kill it with a major change (following semantic versioning).

Also, can you please update the [documentation](https://github.com/adomokos/light-service/blob/master/README.md) with these two helper methods?

.gitignore
```diff
@@ -17,3 +17,4 @@ test/version_tmp
 tmp
 vendor/bundle
 bin
+.idea
```
adomokos: You don't need to add this to `.gitignore`. You can make it "personally ignored" in `.git/info/exclude`. Thanks!

gee-forr:
Hi @adomokos - I’m on board with this. I’m happy to refactor my code to address any LS depreciations that might pop up, so if you wanna deprecate aliases, or even swap em out for what’s here, I’m on board :)

@ni3t - nice PR :)

adomokos:
@ni3t Can you update the PR with the requested changes? I'd like merge this when ready!

adomokos:
Thank you for your PR, I'll update the README.

PR #174, opened 2019-09-30
Make activesupport a runtime dependency

ramontayag:
We use it in tests but we also use it during runtime

Related to https://github.com/adomokos/light-service/pull/157


adomokos:
Thank you!

PR #204, opened 2021-01-26
Add PHP implementation

gee-forr:
Hi there @adomokos 

I've moved the link to the python implementation to a section near the bottom, and added a reference to the PHP implementation by the very talented @douglasgreyling

I ~~brainwashed~~ taught him about lightservice a coupla years ago. He switched jobs to a PHP house, and couldn't live without it. So, he did what any enterprising dev would do, and built his own.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,11 +14,7 @@
 
 <br>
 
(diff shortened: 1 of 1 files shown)


gee-forr:
go home @codecov-io, you're drunk :)

Issue #188, opened 2020-04-27
Rollback Block Not Invoked In Iterated Actions


redjoker011:
Hi I'm a big fan of LightService and been using this incredible gem for years, though upon creating a complex `Organizer` with multiple actions(combination of orchestrator methods such as `reduce_if`, `iterate` and many more functions) A colleague of mine notice that `rolled_back` block is not called on actions inside `:iterate` block(not sure with other helper methods from orchestrator) which based on docs should be invoked, Im not sure if this is a known issue or this is the intended behaviour ?

I've cloned the master branch and try to add test suite with `LightService::Organizer#iterate` below and the test failed to revert the `number` to zero which indicates that the `rolled_back` block inside `LightService::Organizer#iterate` has never called.

Just to add we're using the latest gem version, I hope This issue would be fixed and Thank You for this great gem.

```ruby
# https://github.com/adomokos/light-service/blob/master/spec/acceptance/rollback_spec.rb
require 'spec_helper'

class RollbackWithIterationOrganizer
  extend  LightService::Organizer

  def self.call(number, times)
    with(number: number, times: 0..times).reduce(
      AddsOneWithRollbackAction,
      iterate(:times, AddsOneActionWithRollbackToRevertToZero),
      AddsTwoActionWithRollback,
    )
  end
end

class AddsOneActionWithRollbackToRevertToZero
  extend LightService::Action
  expects :number, :times
  promises :number

  executed do |ctx|
    ctx.number = ctx.number + 1
  end

  rolled_back do |ctx|
    # Block should be invoked
    ctx.number = 0
  end
end

class AddsOneWithRollbackAction
  extend LightService::Action
  expects :number
  promises :number

  executed do |context|
    context.fail_with_rollback! if context.number.zero?

    context.number += 1
  end

  rolled_back do |context|
    context.number -= 1
  end
end

class AddsTwoActionWithRollback
  extend LightService::Action
  expects :number

  executed do |context|
    context.number = context.number + 2
    context.fail_with_rollback!("I did not like this a bit!")
  end

  rolled_back do |context|
    context.number -= 2
  end
end

describe "Rolling back actions when there is a failure" do
  it "Increment using iterate and rollback to revert value to zero" do
    result = RollbackWithIterationOrganizer.call(1, 3)
    number = result.fetch(:number)

    expect(result).to be_failure
    expect(result.message).to eq("I did not like this a bit!")
    expect(number).to eq(0) # FAIL Should revert value to zero
  end
end
```

[...]

PR #205, opened 2021-02-28
Implement rollaback on iterated actions

pioneerskies:
Working on https://github.com/adomokos/light-service/issues/188

### Facts

When in an `Organizer` a `iterate` method is used and if a rollback is called lower in the action chain, an error will throw

![](https://user-images.githubusercontent.com/22144212/80380864-b76b6580-88d2-11ea-9e90-e60faccbe912.png)

### Hypothesis

My opinion is that it happens because while the `Organizer` is reducing actions, the `iterate` method will not return a `LightService::Action`, but a `Proc`, e.g.:

* AddsOneWithRollbackAction <-- Action
* iterate(:times, AddsOneActionWithRollbackToRevertToZero) <-- Proc
* AddsTwoActionWithRollback <-- Action
* 
and the `Proc` wont respond to `rollback`.

### Thesis

Iterated actions should be wrapped in a smart "container" object, not a `Proc`; this way we could make this object able to reply to both `.call` and `.rollback` messages.

The implementation is driven by the goal to use `lib/light-service/organizer/iterate.rb` as entry point and never go higher in the application.

### Implementation

`LightService::Organizer::Iterate` was turned into a `Struct`; it has the really same public interface, but now receiving `.run` it will self-instantiate saving `organizer`, `collection` and `steps` (the iterated action/s) in its internal state.
The instance is returned replacing the previously used `Proc`.

The instance has 2 public methods: `#call` and `#rollback`. Both pretends `context` injected as an argument. This is really important since we cannot save `context` inside the iternal state of any object, but we have to keep a "functional" approach over it within the whole process.

### Tests

A couple of tests has been added

```
 bin/rspec spec/acceptance/rollback_iteration_spec.rb --format=documentation                                                                                                                                                                                                                                           892ms

Randomized with seed 56251

Rolling back actions when there is a failure
  Increment using iterate and rollback to revert value to zero
  when the order of the iterated collection is important
    will apply rollback on each action with in correct order

Finished in 0.00335 seconds (files took 0.48605 seconds to load)
2 examples, 0 failures

Randomized with seed 56251
```


pioneerskies:
Now that I'm reading my own PR description:

```
Rolling back actions when there is a failure
  Increment using iterate and rollback to revert value to zero
  when the order of the iterated collection is important
    will apply rollback on each action with in correct order
```

are anything but explanatory 😂 I'll wait for the review, but let's remember to make those texts better.

adomokos:
Thank you for taking a stab at this. I'll try to review it today or tomorrow. 👍

adomokos:
@pioneerskies, I grabbed your branch and getting this error:

```
$: be rspec spec

An error occurred while loading ./spec/acceptance/rollback_spec.rb.
Failure/Error: raise msg if respond_to?(:rollback)

RuntimeError:
  `rolled_back` macro can not be invoked again
# ./lib/light-service/action.rb:55:in `rolled_back'
# ./spec/acceptance/rollback_spec.rb:92:in `<class:AddsTwoActionWithRollback>'
# ./spec/acceptance/rollback_spec.rb:82:in `<top (required)>'


Finished in 0.00004 seconds (files took 1.95 seconds to load)
0 examples, 0 failures, 1 error occurred outside of examples
```

Do you have that error as well?

pioneerskies:
> @pioneerskies, I grabbed your branch and getting this error:
> Do you have that error as well?

Sure. That error directly points us to an implementation problem. I think that it should be the major next focus if the thesis of the PR could be considered as reasonable.

If you had take a look at why the error is dropped, it should be clear that it is a "business" problem: if a "rollback" can't be called more then once on the same action, but we have the feature to iterate on actions then I think we've a business collision.

What's you pov?

adomokos:
I dived into the code a bit more and you're taking the steps I would take. I like that you are creating a new object for each iteration cycle.

If I built this, I would try to see how I could reuse the `reduce_rollback` from [WithReducer](https://github.com/adomokos/light-service/blob/8888bbfacf1c5e396d26e1cb5224672bbdc6b4af/lib/light-service/organizer/with_reducer.rb#L47-L57). I wonder if we could store a collection of these created "iteration" objects and keep track of them. When a `FailWithRollbackError` is raised, we can just use the collection as [reversable_actions](https://github.com/adomokos/light-service/blob/8888bbfacf1c5e396d26e1cb5224672bbdc6b4af/lib/light-service/organizer/with_reducer.rb#L48) and reuse the same logic there.

What do you think about that @pioneerskies?

pioneerskies:
Thanks for your feedback! <3

I still see a problem: the `.rolled_back` macro in `LightService::Action` module is happening at the `Action` level. The action doesn't know if it's getting called in the context of an iteration or not. Thus it will continue to `raise` regardless of the path we'll take on the upper level.

Could we raise custom exception instead of a generic one and `rescue` it while inside an `Iterate` object? I mean: the `Iterate` object knows that it's ok (is it?) to continue even if that error was raised.

I'd like to hear back from you about this specific problem before to head over on your really stimulating idea.

And about your idea: are you someway imagining the actual `Iterate` as a specialized organizer-in-the-organizer?

lib/light-service/organizer/iterate.rb
```diff
@@ -1,21 +1,39 @@
...
-          item_key = collection_key.to_s.singularize.to_sym
-          collection.each do |item|
-            ctx[item_key] = item
-            ctx = scoped_reduce(organizer, ctx, steps)
+      def rollback(ctx)
+        collection = ctx[collection_key]
+        item_key = collection_key.to_s.singularize.to_sym
+        collection.reverse_each do |item|
+          ctx[item_key] = item
+          steps.reverse_each do |action|
```
gmgp, line 13:
```suggestion
          [steps].flatten.reverse_each do |action|
```

my 2 cent

from readme iterate allow `iterate(:retrieved_items, [TransformsData])`

from my test is possible to use

`iterate(:retrieved_items, TransformsData)`
`iterate(:retrieved_items, [TransformsData])`
`iterate(:retrieved_items, [TransformsData,TransformsData2])`

so the Array refinement can be excluded

gmgp:
with my fix and @pioneerskies code

```ruby

class RollbackWithIterationOrganizer

  def self.call(number, times)
    with(number: number, times: 0..times).reduce(
      AddsTenActionWithRollback,
      iterate(
        :times,
        [
          AddsOneActionWithRollback,
          AddsOneThousandActionWithRollback
        ]
      ),
      AddsOneHundredActionWithRollback,
    )
  end
end

class AddsOneActionWithRollback
  extend LightService::Action
  expects :number, :times
  promises :number

  executed do |context|
    context.number = context.number + 1
    puts "+   1 => #{context.number}"
  end

  rolled_back do |context|
    # Block should be invoked
    context.number -= 1
    puts "-   1 => #{context.number}"
  end
end

class AddsOneThousandActionWithRollback
  extend LightService::Action
  expects :number, :times
  promises :number

  executed do |context|
    context.number = context.number + 1000
    puts "+ 1000 => #{context.number}"
  end

  rolled_back do |context|
    # Block should be invoked
    context.number -= 1000
    puts "- 1000 => #{context.number}"
  end
end

class AddsTenActionWithRollback
  extend LightService::Action
  expects :number
  promises :number

  executed do |context|
    context.fail_with_rollback! if context.number.zero?

    context.number += 10
    puts "+  10 => #{context.number}"
  end

  rolled_back do |context|
    context.number -= 10
    puts "-  10 => #{context.number}"
  end
end

class AddsOneHundredActionWithRollback
  extend LightService::Action
  expects :number

  executed do |context|
    context.number = context.number + 100
    puts "+  100 => #{context.number}"
    context.fail_with_rollback!("I did not like this a bit!")
  end

  rolled_back do |context|
    context.number -= 100
    puts "-  100 => #{context.number}"
  end
end

RollbackWithIterationOrganizer.call(1, 3)

```
produce this output  to help to understand the flux
```
+  10 => 11
+   1 => 12
+ 1000 => 1012
+   1 => 1013
+ 1000 => 2013
+   1 => 2014
+ 1000 => 3014
+   1 => 3015
+ 1000 => 4015
+  100 => 4115
-  100 => 4015
- 1000 => 3015
-   1 => 3014
- 1000 => 2014
-   1 => 2013
- 1000 => 1013
-   1 => 1012
- 1000 => 12
-   1 => 11
-  10 => 1
```

pioneerskies:
Hello there! Just an heartbeat: I'm not out from this one! I'm encountering some difficulties in the design process, moreover since I'm relegating it to spare free time; I feel like this deserves more focus-time instead. I'll try to schedule some of my work training-dedicated time on this matter in the hope to sort something out :)

adomokos:
@pioneerskies - sorry for letting this slide on my end. I meant to set aside some time to think about this and never did that. Thanks for pinging me about it!
I'll play with the code and will provide more pointers by end of this week if that works. Thanks for your patience!!

adomokos:
@pioneerskies - I took a stab at this today, see [my PR](#206). I would love to work with you on it, but the easiest thing for me was to branch off of `master`.

pioneerskies:
Premise:

* I changed the naming a bit (step/steps above all); I needed for my reasonings but idk if it could make sense for you
* I put a lot of classes into the same file; it's just temporary, just messy, just to take the time to understand together where to put things and only if they do make sense
* I changed an expectation in tests (`spec/acceptance/organizer/iterate_spec.rb:27`) because I changed the actual behaviour
* all specs are green here
* taking you specs from #206 they are green too
* my commit history is just a bunch of WIP atm; forgive me but time is always evil (it's still a draft after all)
* I pushed this code after #206 but I already had it locally before reading your; I'll take your other PR 100% into account as soon as I'll have the time to familiarize a bit with it :)

lib/light-service/organizer/iterate.rb
```diff
@@ -1,22 +1,82 @@
...
+          SubOrganizer.new(self).with(ctx).reduce_rollback(single_iteration.actions)
+        end
+
+        ctx
+      end
+
+      def call(ctx)
+        return ctx if ctx.stop_processing?
+
+        setup_iterations(ctx, @collection_key, @actions)
```
pioneerskies, line 36: This is the first time we have the opportunity to read the `context` from the outside, so we setup iterations map here. From now on we'll work on Iterations.

lib/light-service/organizer/iterate.rb
```diff
@@ -1,22 +1,82 @@
...
-        lambda do |ctx|
-          return ctx if ctx.stop_processing?
+      def reversable_iterations
+        index_of_current_iteration = @iterations.members.index(@current_iteration) || 0
 
+        # Reverse from the point where the fail was triggered
+        @iterations.members.take(index_of_current_iteration + 1)
+      end
+
+      class Iterations
```
pioneerskies, line 64: This is the exploded and static representation of the whole iterations. Its `members` is and array of `SingleIteration`. Each `SingleIteration` brings its `item`, `item_key` (probably useless) and the whole set of `Action`s to be run within it.

lib/light-service/organizer/iterate.rb
```diff
@@ -1,22 +1,82 @@
+# require 'light-service/action.rb'
+
 module LightService
   module Organizer
-    class Iterate
-      extend ScopedReducable
+    class Iterator
```
pioneerskies, line 5: I changed the class but not the filename. My fault

lib/light-service/organizer/iterate.rb
```diff
@@ -1,22 +1,82 @@
...
+      private
+
+      def setup_iterations(ctx, collection_key, actions)
+        @iterations ||= Iterations.new(ctx, collection_key, actions)
+      end
 
-      def self.run(organizer, collection_key, steps)
-        lambda do |ctx|
-          return ctx if ctx.stop_processing?
+      def reversable_iterations
```
pioneerskies, line 57: It's really the same concept as per `reverable_actions`, but here is on `SingleIteration` objects

lib/light-service/organizer/iterate.rb
```diff
@@ -1,22 +1,82 @@
...
+
+      def call(ctx)
+        return ctx if ctx.stop_processing?
+
+        setup_iterations(ctx, @collection_key, @actions)
+
+        @iterations.members.each do |single_iteration|
+          ctx[single_iteration.item_key] = single_iteration.item
+          @current_iteration = single_iteration
+          ctx = SubOrganizer.new(self).with(ctx).reduce(single_iteration.actions)
```
pioneerskies, line 41: Note that the `Suborganizer` is getting `self` as the monitored orgnizer. This is leading for sure to some drawbacks like logging or something else I'm not aware. This should be matter of discussion.

lib/light-service/organizer/iterate.rb
```diff
@@ -1,22 +1,82 @@
...
+        end
+
+        ctx
+      end
+
+      def self.run(_organizer, collection_key, actions)
```
pioneerskies, line 47: Preserving the outer interface, but atm `_organizer` is not getting used.

lib/light-service/organizer/iterate.rb
```diff
@@ -1,22 +1,82 @@
...
+        @actions = actions
+        @iterations = nil
+        @current_iteration = nil
+      end
+
+      def rollback(ctx)
+        reversable_iterations.reverse.each_with_index do |single_iteration, idx|
+          ctx[single_iteration.item_key] = single_iteration.item
+
+          unless single_iteration.actions.include?(ctx.current_action)
```
pioneerskies, line 21: `single_iteration.actions.include?(ctx.current_action)` alias "Was the whole process interrupted while inside the iterator?"

lib/light-service/organizer/iterate.rb
```diff
@@ -1,22 +1,82 @@
...
+            ctx.current_action = single_iteration.actions.last
+          end
+
+          ctx.current_action = single_iteration.actions.last unless idx.zero?
```
pioneerskies, line 25: `idx.zero?` alias "Am I on the first interrupted `SingleIteration`?"

pioneerskies: Going to give an exmplanation of `single_iteration.actions.include?(ctx.current_action)` and `idx.zero?`

Having this iteration where numbers are `item` and letters are `Action`s

```
1 [ A B C ]
2 [ A B C ]
3 [ A B C ]
```

failing with rollback on `2B`, thus having `B` as `current_action` in `context`.

Start the rollback for `item` 2:

* "Was the whole process interrupted while inside the iterator?" yes; thus preserve the `current_action`
* "Am I on the first interrupted SingleIteration?" yes I am; thus preserve the `current_action`

The `SubOrganizer` will rollback `2B` then `2A`.

Next cycle will be on item `1`:

* "Was the whole process interrupted while inside the iterator?" yes; thus preserve the `current_action`
* * "Am I on the first interrupted SingleIteration?" no, I am on the second (reversed order), so reset the `current_action` to the first (reversed order) `Action` of this `SingleIteration`

The `SubOrganizer` will rollback `1C` then `1B` then `1A`.

lib/light-service/organizer/with_reducer.rb
```diff
@@ -36,7 +36,11 @@ def reduce(*actions)
           begin
             invoke_action(current_context, action)
           rescue FailWithRollbackError
-            reduce_rollback(actions)
+            if @organizer.is_a? LightService::Organizer::Iterator
+              @organizer.rollback(context)
```
pioneerskies, line 40: Here's one of the main trick of my proposal. Waiting for incoming drawbacks ;)

spec/acceptance/rollback_iteration_spec.rb
```diff
@@ -0,0 +1,172 @@
...
+
+  class CallRollback
+    extend LightService::Action
+
+    executed do |context|
+      context.fail_with_rollback!('Arbitrary rollback.')
+    end
+  end
+
+  class CallRollbackIfSpam
```
pioneerskies, line 70: This is awful because this fixture assumes to find the word `spam` which is indeed passed as params by the test...I'll rethink it, but atm it's doing its job.

lib/light-service/organizer/iterate.rb
pioneerskies: Should I switch back to the original `current_action`? 🤷

lib/light-service/organizer/iterate.rb
```diff
@@ -1,22 +1,80 @@
...
           collection = ctx[collection_key]
           item_key = collection_key.to_s.singularize.to_sym
-          collection.each do |item|
-            ctx[item_key] = item
-            ctx = scoped_reduce(organizer, ctx, steps)
-          end
 
-          ctx
+          @members = collection.each_with_object([]) do |item, matrix|
+            matrix << SingleIteration.new(item, item_key, actions)
```
pioneerskies, line 70: `matrix` is just a reminder from previous code; to be changed.

pioneerskies:
In the meantime I've "merged" your specs from #206 and from this PR into a single ordered file. This way is easier - at least for me - to switch from one branch to the other and debugging running the same specs in both scenarios :)

lib/light-service/organizer/iterate.rb
pioneerskies: Fixed

lib/light-service/organizer/iterate.rb
pioneerskies: Code changed, but the reasoning is still useful. I've added some comment inline to clarify most obsure logic

lib/light-service/organizer/iterate.rb
pioneerskies: fixed

lib/light-service/organizer/iterate.rb
pioneerskies: Nevermind; the code has comments. Let's cleanup the PR

pioneerskies:
Hey @adomokos , I come back onto thi one.

Last time I worked on this I studied your #206 , I've implemented more and stricter specs on it and specs did not pass.
I also tried to refactor you code to turn specs green, but I ended implementing more and more logic too similar to the one in my own PR. Thus I decied that my work was worth to be officially valuated because actually it seems to work.

I've cleaned up, reordered and commented the code. I've also cleaned up commit history.

I'm quite sure to have full test coverage on new code, but unfortunately I'm not able to let codecov understand that I've tested the `refinement` too... Codecov is still complaining; I'm sorry about that and open to any advice on how to improve.

I'll appreciate your feedback :)

Cheers

pioneerskies:
Unfortunately I've later noticed, having had some doubts, that my tests were a golden cage.

With this code, all works ONLY if we have not earlier actions before the iteration. Having earllier actions they won't be rolled back because `Iterator` once consumed its "internal" `rollback` will be unable to instruct the outer organizer to take over.

It's totally not worth to spend a lot of time on reviewing it, thus I'm going to clsoe. 🤷 Maybe one day I'll find the way to innest the sane iteration rollback logic into the other PR.

adomokos:
Thank you for closing it, but I'd still like to see this implemented. 👍

pioneerskies:
> Thank you for closing it, but I'd still like to see this implemented. 👍

In spare time I'm rewriting and elucubrating something else, but still no public branches tho

PR #206, opened 2021-03-20
Iterate with Rollback

adomokos:
We can use this PR or use #205 for it.

This one passes all unit tests but had to comment out a guard for calling a macro twice. Still 👀 .

### DIFF
--- a/lib/light-service/organizer/iterate.rb
+++ b/lib/light-service/organizer/iterate.rb
@@ -3,6 +3,7 @@ module LightService
     class Iterate
       extend ScopedReducable
 
+      # rubocop:disable Metrics/MethodLength
       def self.run(organizer, collection_key, steps)
         lambda do |ctx|
           return ctx if ctx.stop_processing?
@@ -13,19 +14,27 @@ module LightService
             ctx[item_key] = item
 
             ctx = scoped_reduce(organizer, ctx, steps)
-            if (ctx.key?(:_rollback))
-              reversed_processed_collection = collection.take_while { |i| i != item }.reverse
-              rollback_items(organizer, ctx, reversed_processed_collection, steps)
-              break
-            end
+
+            # Handle Rollback
+            next unless ctx.key?(:_rollback)
+
+            reversed_processed_collection = \
+              collection.take_while { |i| i != item }.reverse
+            rollback_items(
+              organizer,
+              ctx,
+              reversed_processed_collection,
+              steps
+            )
           end
 
           ctx
         end
       end
+      # rubocop:enable Metrics/MethodLength
 
       def self.rollback_items(organizer, ctx, collection, steps)
-        collection.each do |item|
+        collection.each do
           ctx = scoped_reduce_rollback(organizer, ctx, steps)
         end
       end
--- a/lib/light-service/organizer/iterate.rb
+++ b/lib/light-service/organizer/iterate.rb
@@ -15,15 +15,14 @@ module LightService
 
             ctx = scoped_reduce(organizer, ctx, steps)
 
-            # Handle Rollback
             next unless ctx.key?(:_rollback)
 
-            reversed_processed_collection = \
-              collection.take_while { |i| i != item }.reverse
+            # Handle Rollback
             rollback_items(
+              item,
               organizer,
               ctx,
-              reversed_processed_collection,
+              collection,
               steps
             )
           end
@@ -33,9 +32,16 @@ module LightService
       end
       # rubocop:enable Metrics/MethodLength
 
-      def self.rollback_items(organizer, ctx, collection, steps)
-        collection.each do
-          ctx = scoped_reduce_rollback(organizer, ctx, steps)
+      def self.rollback_items(item, organizer, ctx, collection, steps)
+        reversed_processed_collection = \
+          collection.take_while { |i| i != item }.reverse
+
+        reversed_processed_collection.each do
+          ctx = scoped_reduce_rollback(
+            organizer,
+            ctx,
+            steps
+          )
         end
       end
     end
--- a/lib/light-service/organizer/iterate.rb
+++ b/lib/light-service/organizer/iterate.rb
@@ -11,10 +11,22 @@ module LightService
           item_key = collection_key.to_s.singularize.to_sym
           collection.each do |item|
             ctx[item_key] = item
+
             ctx = scoped_reduce(organizer, ctx, steps)
+            if (ctx.key?(:_rollback))
+              reversed_processed_collection = collection.take_while { |i| i != item }.reverse
+              rollback_items(organizer, ctx, reversed_processed_collection, steps)
+              break
+            end
           end
 
           ctx
+        end
+      end
+
+      def self.rollback_items(organizer, ctx, collection, steps)
+        collection.each do |item|
+          ctx = scoped_reduce_rollback(organizer, ctx, steps)
         end
       end
     end
--- a/lib/light-service/organizer/iterate.rb
+++ b/lib/light-service/organizer/iterate.rb
@@ -36,7 +36,8 @@ module LightService
         reversed_processed_collection = \
           collection.take_while { |i| i != item }.reverse
 
-        reversed_processed_collection.each do
+        reversed_processed_collection.each do |i|
+          puts "Reversing item #{i}"
           ctx = scoped_reduce_rollback(
             organizer,
             ctx,
--- a/lib/light-service/organizer/scoped_reducable.rb
+++ b/lib/light-service/organizer/scoped_reducable.rb
@@ -10,11 +10,7 @@ module LightService
       end
 
       def scoped_reduce_rollback(organizer, ctx, steps)
-        ctx.reset_skip_remaining! unless ctx.failure?
-        ctx = organizer.with(ctx).reduce_rollback(steps)
-        ctx.reset_skip_remaining! unless ctx.failure?
-
-        ctx
+        organizer.with(ctx).reduce_rollback(steps)
       end
     end
   end
--- a/lib/light-service/organizer/scoped_reducable.rb
+++ b/lib/light-service/organizer/scoped_reducable.rb
@@ -8,6 +8,14 @@ module LightService
 
         ctx
       end
+
+      def scoped_reduce_rollback(organizer, ctx, steps)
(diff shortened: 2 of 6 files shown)


spec/acceptance/organizer/iterate_spec.rb
```diff
@@ -1,44 +1,111 @@
...
-    expect(result.number).to eq(5)
+    expect(result.number).to eq(7)
+  end
+
+  it 'rolls back the actions when it reaches 10' do
+    result = IterateSpec::TestIterate.call(:number => 1,
+                                           :counters => [1, 2, 3, 4])
+
+    expect(result).to be_failure
+    expect(result.number).to eq(1)
```
adomokos, line 79: This is the proof for the logic.

adomokos:
Build is passing on Travis CI, but the code is still a mess. Happy take suggestions on isolating the change. Thanks!

lib/light-service/organizer/scoped_reducable.rb
```diff
@@ -8,6 +8,14 @@ def scoped_reduce(organizer, ctx, steps)
...
         ctx
       end
+
+      def scoped_reduce_rollback(organizer, ctx, steps)
+        ctx.reset_skip_remaining! unless ctx.failure?
+        ctx = organizer.with(ctx).reduce_rollback(steps)
+        ctx.reset_skip_remaining! unless ctx.failure?
+
+        ctx
+      end
```
adomokos, line 18: I wonder if we can use this for other `orchestrator` methods. `reduce_if` and `reduce_until` would need this functionality as well.

lib/light-service/organizer/with_reducer.rb
```diff
@@ -36,6 +36,7 @@ def reduce(*actions)
           begin
             invoke_action(current_context, action)
           rescue FailWithRollbackError
+            context[:_rollback] = true
```
adomokos, line 39: This is bad. But I did not have a better idea to signal for the previous iterator to roll back. I can't reraise the `FailWithRollbackError` as that is internal and what if something won't handle it. I am open to suggestions on how to make this better.

spec/acceptance/organizer/iterate_spec.rb
```diff
@@ -1,44 +1,111 @@
...
+                          FailsWithRollbackWhenReachesTwelve])])
+    end
+
+    def self.call_single(context)
+      with(context)
+        .reduce([iterate(:counters,
+                         AddsOneWithRollbackAction)])
+    end
+  end
+end
```
adomokos, line 61: We already have `AddsOneWithRollbackAction` [here](https://github.com/adomokos/light-service/blob/e291a50f1d49a465fbc1ac82c3b490d426208406/spec/acceptance/rollback_spec.rb#L16-L30), I wonder if these could be merged and put to good use in `spec/test_doubles.rb`.

lib/light-service/organizer/with_reducer_log_decorator.rb
```diff
@@ -50,6 +50,11 @@ def reduce(*actions)
         end
       end
 
+      def reduce_rollback(actions)
+        # TODO: Add logging
+        decorated.reduce_rollback(actions)
+      end
```
adomokos, line 56: I had to expose this function as well.

pioneerskies:
Hey, thank you very much for this :)

I think there is a problem in the proof itself; it is based on the assumption that the rollback is happening on the last action of the last iteration.

Checking out your branch and running spec with this modified Organizer

```ruby
class TestIterate

    def self.call(context)
      with(context)
        .reduce([iterate(:counters,
                         [AddsOneWithRollbackAction,
                          AddsTwoWithRollbackAction,
                          FailsWithRollbackWhenReachesSix, # Note `Six`, not `Twelve`
                          ])])
    end

    def self.call_single(context)
      with(context)
        .reduce([iterate(:counters,
                         AddsOneWithRollbackAction)])
    end
end
```

is breaking things. I think this is happening because the rollback is called before the end of one iteration and the logic is not built to keep it up. This is the operation getting printed from my example w/ current code:

```
"1 + 1"
"2 + 2"
"4 + 1"
"5 + 2"
"7 - 2"
"5 - 1"
"4 - 2"
"2 - 1"
"1 - 2"
"-1 - 1"
"-2 - 2"
"-4 - 1"
"-5 - 2"
"-7 - 1"
"-8 - 2"
"-10 - 1"
"-11 - 2"
"-13 - 1"
```

Your approach is really different from mine - and much cleaner - thus I'll have to take some time to read and debug it more :)

adomokos:
Thanks for pointing out the incorrect logic.
I added one more action to the organizer after the rollback could happens, here is my assumption of what the outcome should be:

```
1st cycle
1 + 1 = 2
2 + 2 = 4
4 + 1 = 5

2nd cycle
5 + 1  = 6
6 + 2 = 8
8 + 1 = 9

3rd cycle
9 + 1 = 10
10 + 2 = 12

-- Starts Rollback ---

3rd cycle
12 - 2 = 10
10 - 1 = 9

2nd cycle
9 - 1 = 8
8 - 2 = 6
6 - 1 = 5

1st cycle
5 - 1 = 4
4 - 2 = 2
2 - 1 = 1
```

Unfortunately, that's not exactly what's happening:

```
Randomized with seed 42574
1 + 1
2 + 2
4 + 1
5 + 1
6 + 2
8 + 1
9 + 1
10 + 2
12 - 2
10 - 1
Reversing item 2
9 - 2
7 - 1
Reversing item 1
6 - 2
4 - 1
Reversing item 3
3 - 2
1 - 1
Reversing item 2
0 - 2
-2 - 1
Reversing item 1
-3 - 2
-5 - 1
F

Failures:

  1) LightService::Organizer rolls back the actions when it reaches 12
     Failure/Error: expect(result.number).to eq(1)

       expected: 1
            got: -6
```

I suspect this could be fixed, but I don't have time to look at it right now or the remainder of the week. Sorry.

If I were you, and you like the logic I've coded so far, I'd say just start a new branch from master, cherry-pick my commits I made in this PR, and push this change through the finish line.

I liked that the solution I came up with, the core logic was only changed at two places, all other changes were isolated in the [iterate.rb](https://github.com/adomokos/light-service/blob/7554bff953ba87264d01dac6f6b3d1c95029034f/lib/light-service/organizer/iterate.rb#L20-L28) file.

lib/light-service/organizer/iterate.rb
```diff
@@ -11,12 +12,39 @@ def self.run(organizer, collection_key, steps)
           item_key = collection_key.to_s.singularize.to_sym
           collection.each do |item|
             ctx[item_key] = item
```
pioneerskies, line 14: I think it could be better to "advance" the item only if we're not inside a rollback.
Also I can't get why we're not `break`ing the `each`'s cycle if we're whilst a _rollback_.

E.g.:

```ruby
            if ctx.key?(:_rollback)
              # Handle Rollback
              rollback_items(
                item,
                organizer,
                ctx,
                collection,
                steps
              )
              break
            else
              ctx[item_key] = item
              ctx = scoped_reduce(organizer, ctx, steps)
            end
```

lib/light-service/organizer/iterate.rb
```diff
@@ -11,12 +12,39 @@ def self.run(organizer, collection_key, steps)
...
           end
 
           ctx
         end
       end
+      # rubocop:enable Metrics/MethodLength
+
+      def self.rollback_items(item, organizer, ctx, collection, steps)
+        reversed_processed_collection = \
+          collection.take_while { |i| i != item }.reverse
```
pioneerskies, line 37: Probably the is a sneaky fact happening invalidating this logic approach: when execution is at https://github.com/adomokos/light-service/pull/206/files#diff-0c78fccee9a0c066001d7b16dd8fe4c02ceebbfa897522ec9f9ecac8d7b4f2f7R39 the Organizer is immediately receiving the first `reduce_rollback` taking the last occurred iteration's actions.

Let's suppose to have

```
1 [A B C]
2 [A B C]
```

and to `fail_with_rollback!` at `2B`. The organizer will immediately receive `reduce_rollback(actions)` having `actions = 2 [A B C]` (if you forgive the pseudocode). Thus, when execution will eventually come back to our `Iterate` class (I'll have to write more on this _eventually_ later), we have to consider that `2 [A B C]` was already rolled back.

This is sneaky because it's a fact, but from the pov of `Iterate` is a matter of faith: it happens in a place it has not vision on.

That said my hypothesis is that `collection.take_while { |i| i != item }` is _taking_ the already rolled back "item" too (already taking in consideration my note at https://github.com/adomokos/light-service/pull/206/files#r601165629)

lib/light-service/organizer/iterate.rb
```diff
@@ -11,12 +12,39 @@ def self.run(organizer, collection_key, steps)
           item_key = collection_key.to_s.singularize.to_sym
           collection.each do |item|
             ctx[item_key] = item
+
             ctx = scoped_reduce(organizer, ctx, steps)
+
+            next unless ctx.key?(:_rollback)
+
+            # Handle Rollback
+            rollback_items(
```
pioneerskies, line 21: I'm wondering (still not had the time to translate this wonder into a spec): if we're attaching `Iterate`'s rollback capability (`Iterate.rollback_items`) inside the reduce cycle what would happen if `fail_with_rollback!` will be called on the last iteration or after the `iterate` at all?

My hypothesis is that it will never been called, thus this logic is able to handle rollback only happened within the `collection.count - 1`th iteration cycle. Not on the last one nor if the rollback will be called from a later action outsite the `iterate`.

Am I missing something?

pioneerskies:
I'm having little and spare time: I'm sorry if I'm iteratively (LOL) adding comments day by day.

I could try to sustain my comments with some specs, but I think I won't be able to restart a branch from this one, because I'm observing some problems I previously encountered while thinking and coding for #205. Those problems brought me to do heavier "lifting" and introducing new objects and logics. You've been thin, concise and you're obviously able to better manage  side effects such as preserving correct logging facilities, think ahead of time about code reuse and on and on; but some really core problems are unsolved here, as far as I can understand.

I hope my comments will be of some contribute at least. 😃

adomokos:
I am happy to close this PR, if #205 is closer to the finish line. I figured this PR could be a good way to show you how I'd like to isolate new logic from the "functional core".

adomokos:
Inactive PR, closing it.

PR #209, opened 2021-06-23
🐛 Fix callback current action

jaredsmithse:
I came across a bug where the `current_action` is not changed back correctly after invoking callbacks. This has the consequence of any kind of after action callbacks not having the correct `current_action` set.

For the test that was added, if you add `puts ctx.current_action` to the `after_actions` hook, you'll get the following printout:
```
TestDoubles::IncrementCountAction
TestDoubles::AddToTotalAction
TestDoubles::IncrementCountAction
TestDoubles::AddToTotalAction
TestDoubles::IncrementCountAction
TestDoubles::AddToTotalAction
TestDoubles::AddToTotalAction
```

You'll see the last action is printed twice. That's because the very last one should have `TestDoubles::IterateCollectionAction` as it's `:current_action` instead of a duplicate `TestDoubles::AddToTotalAction`. 

I took a stab at adding the test and where I think is the simplest place to make it pass. After the change to `action.rb`, the output from that same `puts` statement looks like:

```
TestDoubles::SetUpContextAction
TestDoubles::IncrementCountAction
TestDoubles::AddToTotalAction
TestDoubles::IncrementCountAction
TestDoubles::AddToTotalAction
TestDoubles::IncrementCountAction
TestDoubles::AddToTotalAction
TestDoubles::IterateCollectionAction
```

This gets the desired effect for setting the current action, let me know if there is a better way to go about solving this! It's also my first time contributing to OS in a while, so any other feedback is welcome 🙏 

[Gitmoji Ref](https://gitmoji.dev/)

### DIFF
--- a/lib/light-service/action.rb
+++ b/lib/light-service/action.rb
@@ -44,6 +44,8 @@ module LightService
             catch(:jump_when_failed) do
               call_before_action(action_context)
               yield(action_context)
+
+              # Reset the stored action in case it was changed downstream
               action_context.current_action = self
               call_after_action(action_context)
             end
--- a/lib/light-service/action.rb
+++ b/lib/light-service/action.rb
@@ -44,6 +44,7 @@ module LightService
             catch(:jump_when_failed) do
               call_before_action(action_context)
               yield(action_context)
+              action_context.current_action = self
               call_after_action(action_context)
             end
           end
--- a/spec/acceptance/after_actions_spec.rb
+++ b/spec/acceptance/after_actions_spec.rb
@@ -65,6 +65,23 @@ RSpec.describe 'Action after_actions' do
     end
   end
 
+  context 'with callbacks' do
+    it 'ensures the correct :current_action is set' do
+      TestDoubles::TestWithCallback.after_actions = [
+        lambda do |ctx|
+          if ctx.current_action == TestDoubles::IterateCollectionAction
+            ctx.total -= 1000
+          end
+        end
+      ]
+
+      result = TestDoubles::TestWithCallback.call
+
+      expect(result.counter).to eq(3)
+      expect(result.total).to eq(-994)
+    end
+  end
+
   describe 'after_actions can be appended' do
     it 'adds to the :_after_actions collection' do
       TestDoubles::AdditionOrganizer.append_after_actions(


adomokos:
This is nice! I'll review it today (more deeply). 😍

adomokos:
You stumbled upon a very unique edge case - thank you for it!

The `throw` - `catch` is my poor-man's-version of a GOTO (remember, they are evil) logic to jump from one place in the code to another. An internal publish/subscribe mechanism would be nicer for this, but then again, `LightService` wouldn't be so light anymore. With orchestrators, maybe it's not...

I see another place where we set the `current_action` [here](https://github.com/adomokos/light-service/blob/c87e2c26206a50caa4410bde77f4f18bca02cb33/lib/light-service/action.rb#L39), I wish we could combine that and the one you added into a single one. I am not sure it's possible, I looked, I tweaked your solution, did not work.

Is there any way we could combine those two assignments into one? In case you confirm there is no way to do it, I'll merge this PR.

Thank you so much for adding the failing test first and then putting in the logic to pass it. Absolutely beautiful! 👍

jaredsmithse:
I was thinking this too, it's definitely not ideal 😕 Here's a bit of a dump of my other thoughts:

iirc my first approach was to try and modify `WithCallback` to hold the previous action like we do the callback, but because of how it gets set in the action, we'd have to introduce some notion of `:previous_action` on the context so it has knowledge of it there (and I'm not sure if this has any other side-effects), which seemed like a worse evil, so this came up as my best effort 😅 

Ideally without changing a lot I wish the code could look like this:
```rb
Context::KeyVerifier.verify_keys(action_context, self) do
  action_context.define_accessor_methods_for_keys(all_keys)

  catch(:jump_when_failed) do
    # Store the action within the context
    action_context.current_action = self
    call_before_action(action_context)
    yield(action_context)

    # Reset the stored action in case it was changed downstream
    action_context.current_action = self
    call_after_action(action_context)
  end
end
```

Of course this brings up a new issue because `ReservedKeysVerifier` checks for `:current_action`, and it would be weird to take it out 😓 

Maybe it would help to add a comment like the one in the above example? 😕 🤔

adomokos:
Thank you for your response. Yes, please add the comment to the code and I'll merge your PR. Thank you for it!

adomokos:
Thank you! :100:

jaredsmithse:
Of course, thank you for all of your hard work making it :slightly_smiling_face:

It's been a pleasure using it these past 7+ months and will continue to for quite some time :green_heart:

adomokos:
I'll try to get a release out this weekend.

PR #211, opened 2021-07-02
Context accessors

### LINKED ISSUE #210: Surprising `add_to_context` orchestrator behaviour
Hiya @adomokos !

Got a weird one... I sometimes have organizers that use `iterate` and need to get the promised values from each iteration into something in the context.

Here's a dummy example:

```ruby
# In an imaginary organizer called ApplyRatingsToBook
def self.actions
  [
    add_to_context(book_ratings: []),
    iterate(:ratings, [
      Book::ApplyRating # <- thus guy expects a `Book` and `Rating`, and promises a 'BookRating' for e.g.
      execute(->(ctx) { ctx.book_ratings << ctx.book_rating }) # I'm collecting all `book_ratings` for something outside of the organizer
  ]
end
```

# The surprising behaviour

I would assume that when I add something to the context using `add_to_context` like above, it's available as an accessor in much the same way that `expects` and `promises` works, except it is only accessible as a key on the `Context` object which is ultimately returned from the organizer.

# A way forward

If you agree with the above, I'd love to put together a PR that updates the Organizer `add_to_context` feature to turn it into a context accessor as well as the standard key.

Let me know :) and thanks again :)

gee-forr:
Hi there @adomokos 

Here's a PR that should resolve #210 

Let me know what you think. I also updated the README


README.md
```diff
@@ -7,12 +7,7 @@
 [![License](https://img.shields.io/badge/license-MIT-green.svg)](http://opensource.org/licenses/MIT)
 [![Download Count](https://ruby-gem-downloads-badge.herokuapp.com/light-service?type=total)](https://rubygems.org/gems/light-service)
 
-<br>
-
-![Orchestrators-Deprecated](resources/orchestrators_deprecated.svg)
-<br>Version 0.9.0 deprecates Orchestrators and moves all their functionalities into Organizers. Please check out [this PR](https://github.com/adomokos/light-service/pull/132) to see the changes.
-
-<br>
```
adomokos, line 15: :-)

spec/acceptance/organizer/add_to_context_spec.rb
```diff
@@ -14,6 +14,8 @@ def self.steps
         # with the value of 0, so it's available for
         # AddsOneAction
         add_to_context(:number => 0),
+        add_to_context(:message => "Just a message"),
+        add_to_context(:error_code => "D0H"),
```
adomokos, line 18: I wonder if we should just raise an error when someone tries to set these "reserved" keys via `add_to_context`. What do you think?

adomokos (commented):
Great PR, thank you for it @gee-forr!
Thanks for the README polish. Next time, I'd suggest submitting a separate PR for it to keep it clean.

What do you think about blocking the user from adding values with reserved keys?

spec/acceptance/organizer/add_to_context_spec.rb
gee-forr: I was thinking the exact same thing, and that I should raise an exception, but was concerned about breaking backwards compat.

That being said, if you're injecting a `message` context key using `add_to_context` you're probably doing so with full knowledge that it's not the smartest move.

Also, LS is pre1.0, so breaking changes are not a big deal. Lemme make those changes. Stand by.

gee-forr:
@adomokos - OK - this should be good to go. `add_to_context` will explode if you attempt to set a reserved key now :)

adomokos (approved):
Looks great! Thank you for it!

gee-forr:
Whoop whoop! Nice :) Looking forward to seeing a new ver on rubygems :)

adomokos:
I'll try to get a new version out today.

Issue #215, opened 2021-07-26
How to handle default values for dependency injection?


metalelf0:
Hi, I'm trying out this gem and I'm facing the following issue.

Let's say I have the following action (it's not a real one, it's some code written just to show the issue):

```ruby
class SomeItemExporter
  extend ::LightService::Action

  expects :start_date, :end_date

  executed do |ctx|
    items = ItemRepository.by_date(ctx.start_date, ctx.end_date)
    formatted_items = items.map { |item| ItemPresenter.present(item) }
    csv_builder = ItemToCsv.build_file(formatted_items)
  end
end
```

And let's say that I want to unit test this action, passing in the dependencies via dependency injection.

If I had written the service as a "PORO" it would have looked something like:

```ruby
class SomeItemExporter
  attr_reader :repository, :presenter, :csv_builder

  def initialize(args = {})
    repository  = args[:repository_klass]  || ItemRepository
    presenter   = args[:presenter_klass]   || ItemPresenter
    csv_builder = args[:csv_builder_klass] || ItemToCsv
  end

  def run(start_date, end_date)
    items = repository.by_date(start_date, end_date)
    formatted_items = items.map { |item| presenter.present(item) }
    csv_builder = csv_vuilder.build_file(formatted_items)
  end
end
```

Allowing calls like:

```ruby
# production code, so use defaults
SomeItemExporter.new.run(2.weeks.ago, Date.today)

# test code, so inject stubs
fake_repository = double()
SomeItemExporter.new(repository: fake_repository).run(2.weeks.ago, Date.today)
```

However, doing anything similar in `LightService` would require passing the real classes as `expects` also in production code, making the code a lot more verbose.

Would it be possible to add (or would you be interested in adding) something like this:

```ruby
class SomeItemExporter
  extend ::LightService::Action

  expects :start_date, :end_date
  allows repository: ItemRepository, presenter: ItemPresenter
```
 
So that you could write `ctx.repository` and have either an injected item or the default one?

Hope my point is clear. Let me know if it's not.

pioneerskies:
Said that I'm just a user, and probably I'm raising a more philosophical than a practical argument: given your example I'd go for a PORO. I can't see any advantage here using LightService. You're not using the "functional" approach of using the `context` as single source of truth, neither the flow control facilities to control success/failure from the outside of the service object.

If I'd have to force your example to be a use case for LS (but it's probably a useless excercise)

```ruby
class Format
  extend ::LightService::Action

  expects :items
[...]

PR #216, opened 2021-07-26
✨ Permit allowed attributes

metalelf0:
Allowed attributes can be defined in the action context for dependency
injection. If they aren't explictly passed, the default value is used.

This commit is mostly a proof or concept, more work is needed to add
documentation, more tests etc., but I'd like to receive your opinion about
it before eventually adding more work.

E.g.:

```ruby
  class AddsArgumentOrTwo
    extend LightService::Action
    allows increment: 2
    expects :counter
    promises :result

    executed do |context|
      context.result = context.counter + context.increment
    end
  end
```

### DIFF
--- a/lib/light-service/action.rb
+++ b/lib/light-service/action.rb
@@ -14,12 +14,20 @@ module LightService
     end
 
     module Macros
+      def allows(args)
+        allowed_keys.merge!(args)
+      end
+
       def expects(*args)
         expected_keys.concat(args)
       end
 
       def promises(*args)
         promised_keys.concat(args)
+      end
+
+      def allowed_keys
+        @allowed_keys ||= {}
       end
 
       def expected_keys
@@ -32,6 +40,9 @@ module LightService
 
       def executed
         define_singleton_method :execute do |context = {}|
+          allowed_keys.each do |key, default_value|
+            context[key] ||= default_value
+          end
           action_context = create_action_context(context)
           return action_context if action_context.stop_processing?
 
@@ -73,7 +84,7 @@ module LightService
       end
 
       def all_keys
-        expected_keys + promised_keys
+        expected_keys + promised_keys + allowed_keys.keys
       end
 
       def call_before_action(context)
--- a/spec/action_allowed_keys_spec.rb
+++ b/spec/action_allowed_keys_spec.rb
@@ -1,0 +1,23 @@
+require 'spec_helper'
+require 'test_doubles'
+
+describe ":allows macro" do
+  context "when allowed keys are passed in" do
+    it "uses the passed value" do
+      resulting_context = TestDoubles::AddsArgumentOrTwo.execute(
+        counter: 10,
+        increment: 5
+      )
+      expect(resulting_context[:result]).to eq(15)
+    end
+  end
+
+  context "when allowed keys are not passed in" do
+    it "uses the default value" do
+      resulting_context = TestDoubles::AddsArgumentOrTwo.execute(
+        counter: 10
+      )
+      expect(resulting_context[:result]).to eq(12)
+    end
+  end
+end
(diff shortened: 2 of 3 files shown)


adomokos:
This is cool. I remember we experimented with this, but we called it `optional`. I thinkI might like `allows` better.

adomokos:
Sorry, I fat-fingered the PR, I did not mean to close it. :-)

adomokos:
Please go ahead, if the tests are there and documentation is prepared, I'll merge it. 👍

metalelf0:
Mmm, I'm a bit at a stop point now. The idea works well when testing or executing individual actions. However, without the chance of passing allowed keys in the `reduce` block in the organizers, the whole dependency injections seems to be pretty useless.

What do you think about it? I don't want to overcomplicate things. This is something that *could* possibly work, but I'm not 100% sure it fits the LS philosophy:

```
    Class.new do

      aliases :promised_key => :expected_key

      def self.call(ctx = {})
        with(ctx).reduce(
          [
             [ TestDoubles::AddsArgumentOrTwo, increment: 10 ],
             TestDoubles::AddsArgumentOrTwo,
            # ... any other action
          ]
        )
      end
    end
```

pioneerskies:
Why you think

```ruby
OrganizerClass.call(increment: 10)
```

wouldn't fit the requirement?

metalelf0:
> Why you think
> 
> ```ruby
> OrganizerClass.call(increment: 10)
> ```
> 
> wouldn't fit the requirement?

Cause it would allow only one `increment` per-organizer. In this simple scenario it's perfectly fine, but if I had a complex organizer, where I'm passing e.g. a `repository` to each action, I'd need to use a different name for each repository.

It would be `OrganizerClass.call(first_repository: FirstRepository, second_repository: SecondRepository)` instead of:

```ruby
    Class.new do

      aliases :promised_key => :expected_key

      def self.call(ctx = {})
        with(ctx).reduce(
          [
             [ FirstAction, repository: FirstRepository ],
             [ SecondAction, repository: SecondRepository ],
            # ... any other action
          ]
        )
      end
    end
```

ni3t:
> Mmm, I'm a bit at a stop point now. The idea works well when testing or executing individual actions. However, without the chance of passing allowed keys in the `reduce` block in the organizers, the whole dependency injections seems to be pretty useless.
> 
> What do you think about it? I don't want to overcomplicate things. This is something that _could_ possibly work, but I'm not 100% sure it fits the LS philosophy:
> 
> ```ruby
>     Class.new do
>       extend LightService::Organizer
> 
>       aliases :promised_key => :expected_key
> 
>       def self.call(ctx = {})
>         with(ctx).reduce(
>           [
>              [ TestDoubles::AddsArgumentOrTwo, increment: 10 ],
>              TestDoubles::AddsArgumentOrTwo,
>             # ... any other action
>           ]
>         )
>       end
>     end
> ```

The problem with this interface would be that the `WithReducer` will flatten the nested array when reducing ([here](https://github.com/adomokos/light-service/blob/78d772c078cb7e5aaea55d11fe9aed779c310e66/lib/light-service/organizer/with_reducer.rb#L33)), so passing in the `allows` argument wouldn't work here.

I could see an interface like this working by hacking into the Action class `call` method, but I'm not sure how much the Action class would have to change to support it:

```rb
[
  TestDoubles::AddsArgumentOrTwo[increment: 10],
  TestDoubles::AddsArgumentOrTwo
]
```

gee-forr:
@metalelf0 and @adomokos - I didn't realise you had a PR for this already...  I proposed a slightly different API that makes `expects` take a default static value or proc.

This would avoid [the issue mentioned in this comment](#issuecomment-887371569)

Let me know what you think... an expanded description is here: https://github.com/adomokos/light-service/issues/215#issuecomment-904544844

metalelf0:
This is already done (better) in #219, so I can close this PR. Great work guys!

PR #217, opened 2021-07-26
Fix around_action hook for nested actions

alitsiya:
We need an ability to trace timing for specific actions inside organizers. There is a `around_action` hook. But there is an issue with `around_action` that it's not working for nested Actions that are invoked via proc (example is `reduce_if`). This is an attempt to fix the issue with `around_action`.  See a spec added in `spec/acceptance/around_each_spec.rb`:

```
expect(result[:logger].logs).to eq(
      [
        { :action => TestDoubles::AddsOneAction, :before => 1, :after => 2 },
        { :action => TestDoubles::AddsTwoAction, :before => 2, :after => 4 },
        { :action => TestDoubles::AddsThreeAction, :before => 4, :after => 7 }
      ]
    )
```

Where actual result is 
```
expect(result[:logger].logs).to eq(
      [
        { :action => TestDoubles::AddsOneAction, :before => 1, :after => 2 },
        { :action => TestDoubles::AddsThreeAction, :before => 2, :after => 7 }
      ]
    )
```
So it skips entirely nested actions.

This PR is an attempt to fix issue with `around_action`.

Verified fix using forked version:
before:
Screen Shot 2021-08-26 at 16 16 25

after:
Screen Shot 2021-08-26 at 16 15 59

NOTE:
The solution adds unfortunate side effect of having `_around_actions` key in a context.

### DIFF
--- a/spec/acceptance/log_from_organizer_spec.rb
+++ b/spec/acceptance/log_from_organizer_spec.rb
@@ -61,8 +61,7 @@ describe "Logs from organizer" do
 
     it "lists the keys in context after the actions are executed" do
       organizer_log_message = "[LightService] -     keys in context: " \
-                              ":tea, :milk, :coffee, :_around_actions, " \
-                              ":milk_tea, :latte"
+                              ":tea, :milk, :coffee, :milk_tea, :latte"
       expect(log_message).to include(organizer_log_message)
     end
   end
--- a/spec/acceptance/log_from_organizer_spec.rb
+++ b/spec/acceptance/log_from_organizer_spec.rb
@@ -61,7 +61,8 @@ describe "Logs from organizer" do
 
     it "lists the keys in context after the actions are executed" do
       organizer_log_message = "[LightService] -     keys in context: " \
-                              ":tea, :milk, :coffee, :_around_actions, :milk_tea, :latte"
+                              ":tea, :milk, :coffee, :_around_actions, " \
+                              ":milk_tea, :latte"
       expect(log_message).to include(organizer_log_message)
     end
   end
--- a/spec/acceptance/log_from_organizer_spec.rb
+++ b/spec/acceptance/log_from_organizer_spec.rb
@@ -61,7 +61,7 @@ describe "Logs from organizer" do
 
     it "lists the keys in context after the actions are executed" do
       organizer_log_message = "[LightService] -     keys in context: " \
-                              ":tea, :milk, :coffee, :milk_tea, :latte"
+                              ":tea, :milk, :coffee, :_around_actions, :milk_tea, :latte"
       expect(log_message).to include(organizer_log_message)
     end
   end
--- a/spec/acceptance/log_from_organizer_spec.rb
+++ b/spec/acceptance/log_from_organizer_spec.rb
@@ -59,7 +59,7 @@ describe "Logs from organizer" do
       expect(log_message).to include(organizer_log_message)
     end
 
-    it "lists the keys in contect after the actions are executed" do
+    it "lists the keys in context after the actions are executed" do
       organizer_log_message = "[LightService] -     keys in context: " \
                               ":tea, :milk, :coffee, :milk_tea, :latte"
       expect(log_message).to include(organizer_log_message)
--- a/spec/test_doubles.rb
+++ b/spec/test_doubles.rb
@@ -74,6 +74,16 @@ module TestDoubles
     end
   end
 
+  class AroundEachWithReduceIfOrganizer
+    extend LightService::Organizer
+
+    def self.call(action_arguments)
+      with(action_arguments)
+        .around_each(AroundEachLoggerHandler)
+        .reduce(ReduceIfOrganizer.actions)
+    end
+  end
+
   class AddsTwoActionWithFetch
     extend LightService::Action
 
--- a/lib/light-service/action.rb
+++ b/lib/light-service/action.rb
@@ -45,7 +45,7 @@ module LightService
             catch(:jump_when_failed) do
               call_before_action(action_context)
 
-              if context.instance_of?(Context) and context.around_actions.respond_to?(:call)
+              if around_action_context?(context)
                 context.around_actions.call(action_context) do
                   yield(action_context)
                   action_context
@@ -102,6 +102,10 @@ module LightService
         end
 
         context
+      end
+
+      def around_action_context?(context)
+        context.instance_of?(Context) && context.around_actions.respond_to?(:call)
       end
     end
   end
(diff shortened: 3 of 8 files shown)


adomokos:
Thank you for the draft PR! One thing that I regretted with LightService was introducing its only dependency: `active_support`. Some might say `active_support` should be part of the language. Yeah, sure, but still, it's a dependency that I wanted to avoid.

Can you just use one of the `before_action`/`after_action` hooks to send these messages to Datadog? We added those "hooks" so people can extend it with anything they want.

I still have the dream of yanking `active_support` out of the library, but maybe it's too far-fetched. I am curious to hear what other ppl think.

alitsiya (commented):
> Thank you for the draft PR! One thing that I regretted with LightService was introducing its only dependency: `active_support`. Some might say `active_support` should be part of the language. Yeah, sure, but still, it's a dependency that I wanted to avoid.
> 
> Can you just use one of the `before_action`/`after_action` hooks to send these messages to Datadog? We added those "hooks" so people can extend it with anything they want.
> 
> I still have the dream of yanking `active_support` out of the library, but maybe it's too far-fetched. I am curious to hear what other ppl think.

Thanks, Attila, for your quick reply. 

We want to have ability to trace single actions without organizers, but  around_action` and before/after action hooks can be used with organizers. 
Also, `around_action` is not reliable when actions are nested in `reduce_if` context (added [a spec ](https://github.com/adomokos/light-service/pull/217/files#diff-2faf87bc8d92cbeb7326133e5ad3e80750fed2a6744219e7980a2a146cb74382R20-R30) that shows this case)

adomokos:
Should we fix `reduce_if` then?

alitsiya:
We looked at a potential solution for `around_action` with nested actions and here is what we have to far https://github.com/adomokos/light-service/compare/main...jaredsmithse:jps-around-each-options (also copied to this PR).

adomokos:
Thank you, I'll take a look at this over the weekend.

adomokos:
I did take a look.  It's a bit hard for me to understand the purpose of this PR.

Your first sentence in the first comment might provide some clue:

> Add ability to do instrumentation around actions via ActiveSupport::Notifications without adding unnecessary latency.

How big is this latency? How much better the code is going to get with this PR?

In case you want to use AS Notification, I am good with "open for extension and closed for modification"-kinda change, providing hooks where you can "plug in" your own, custom code.

Can you please create an acceptance test that demonstrates what this PR will add to LS? I am happy to merge it in:
* if it adds value
* fixes bugs
* makes the code cleaner

~ which I think it will do.

Thank you so much and let's get this PR rollin'!

alitsiya:
Hi @adomokos , thanks for looking into this and sorry for not modifying the description after second iteration. I'm not necessary advocating for adding modifications for a 'hook up', since there is an `around_action`.
 It's an attempt to fix a bug where nested around_actions are not invoked. So I updated the description now from `ActiveSupport::Notifications` to a bug fix.

> I did take a look. It's a bit hard for me to understand the purpose of this PR.
> 
> Your first sentence in the first comment might provide some clue:
> 
> > Add ability to do instrumentation around actions via ActiveSupport::Notifications without adding unnecessary latency.
> 
> How big is this latency? How much better the code is going to get with this PR?

It's a pure guess. I haven't measured the impact. 

> 
> In case you want to use AS Notification, I am good with "open for extension and closed for modification"-kinda change, providing hooks where you can "plug in" your own, custom code.
> 
> Can you please create an acceptance test that demonstrates what this PR will add to LS? I am happy to merge it in:
> 
> * if it adds value
> * fixes bugs
> * makes the code cleaner
> 
> ~ which I think it will do.
> 
> Thank you so much and let's get this PR rollin'!

adomokos:
Thank you for your work/time investment into fixing this! I am looking forward to seeing this PR completed!

spec/acceptance/log_from_organizer_spec.rb
```diff
@@ -59,9 +59,9 @@ def collects_log
       expect(log_message).to include(organizer_log_message)
     end
 
-    it "lists the keys in contect after the actions are executed" do
+    it "lists the keys in context after the actions are executed" do
```
adomokos, line 62: 👍  Thank you!

spec/acceptance/log_from_organizer_spec.rb
```diff
@@ -59,9 +59,9 @@ def collects_log
...
       organizer_log_message = "[LightService] -     keys in context: " \
-                              ":tea, :milk, :coffee, :milk_tea, :latte"
+                              ":tea, :milk, :coffee, :_around_actions, :milk_tea, :latte"
```
adomokos, line 64: This is internal to LS, we should not have this visible in the `Context`. What do you think about adding this field to the `Context`? Like we do with `current_action`, [here](https://github.com/adomokos/light-service/blob/0a70d929e6f0b1465ba3cfaede74511f39a125db/lib/light-service/context.rb#L11).

spec/test_doubles.rb
```diff
@@ -74,6 +74,16 @@ def self.call(action_arguments)
...
 
+  class AroundEachWithReduceIfOrganizer
+    extend LightService::Organizer
+
+    def self.call(action_arguments)
+      with(action_arguments)
+        .around_each(AroundEachLoggerHandler)
+        .reduce(ReduceIfOrganizer.actions)
+    end
+  end
```
adomokos, line 85: Thank you for adding this!

adomokos (commented):
Looks really good, I'll merge this. Can you please make the `_around_actions` hidden from the context keys? I think we need to add it to the `Context` as we did it with `current_action`.

spec/acceptance/log_from_organizer_spec.rb
alitsiya: Agree, `_around_actions` is internal. 
I'm not sure I understand your phrase and a little confused: if we explicitly add it to `attr_accessor` it's a public attribute. But you also saying that we need to hide it. Do you have an example in mind how to achieve invisibility?

adomokos:
@alitsiya - can you please get the CI build green? It seems rubocop needs a bit of TLC.

adomokos:
@alitsiya - what do you think about [this](https://github.com/adomokos/light-service/commit/58422720d3de05b7326e8eeb8934cd17ff3ae525)?

It passes all the tests, but Rubocop is still failing.

We have `current_action` as an attribute on the `Context`, `around_actions` is similar to that. Is this the best solution? I don't think so, but I'd prefer this over having a special key in the Context that the developer (the library user) is exposed to.

In my solution I am not a fan of this line:

```
if context.instance_of?(Context) and context.around_actions.respond_to?(:call)
```

If you have ideas to make this more elegant, I am all for that.

Let me know what you think.

alitsiya:
This looks great! Thanks for the help! Cherry-picked your suggestions!

adomokos:
@alitsiya - this is ready to be merge. Can you please fix `rubocop`? Thank you!

adomokos:
Thank you for working on this @alitsiya! 💯

PR #218, opened 2021-08-18
Add ReduceIfElse macro

ni3t:
This PR adds an organizer macro called `reduce_if_else`

The basic idea is similar to the existing `reduce_if`, but adds a third argument that allows the developer to execute a series of steps in the case that the condition block evaluates to false.

For example:

```rb
class AvengersInfinityWar
  extend ::LightService::Organizer

  def self.call(avenger:)
    with(avenger: avenger).reduce(actions)
  end

  def self.actions
    [
      FindInfinityStone,
      reduce_if_else(
         ->(ctx) { ctx[:avenger] == "Hulk" },
        [ HulkSmash ],
        [ KeepAwayFromThanos ]
      )
    ]
  end
end
```

If the context above is called with `{avenger: "Thor"}`, then `KeepAwayFromThanos` will execute. If the service is called with `{avenger: "Hulk"}`, he will try to smash the infinity stone.


adomokos:
I like this! Will try to review it by end of the week. Thank you for it!

adomokos:
I'll tweak the `acceptance` test, but this looks great! Thank you so much for adding this @ni3t! 👍

adomokos:
I wasn't sure if the `if_steps` or `else_steps` accept a single action or it has to be a collection (list).

I was going to change the acceptance test like this:

```
    class TestReduceIfElse

    def self.call(context)
      with(context).reduce(actions)
    end

    def self.actions
      [
        TestDoubles::AddsOneAction,
        reduce_if_else(
          ->(ctx) { ctx.number == 1 },
          TestDoubles::AddsOneAction,
          [TestDoubles::AddsOneAction, TestDoubles::AddsTwoAction]
        )
      ]
    end
```

but I decided against it: I did not want to make the test more complex than it is.

PR #220, opened 2021-08-29
Extract function from executed macro

adomokos:
Rubocop methodlength error is a good indication that the encapsulated
logic in a method is too much. The first step of understanding the
dependencies is to extract logic into a function, this is what I am
doing with this commit.


adomokos:
// cc @alitsiya and @gee-forr.

PR #222, opened 2021-09-07
Add light-service.js by @douglasgreyling to implementations in README

gee-forr:
Add another implementation to the list


adomokos:
Thank you!

PR #224, opened 2021-09-23
Support for named argument in ruby 3

JeanSebTr:
I've tried to update one of our project to ruby 3.0 and the only issue was in our services spec using ContextFactory.

We use named argument a lot and 3.0 break this when they're not passed explicitly using the double splat operator. `ruby2_keywords` marks the method as using the old style of named arguments. Calling it this way should works with older 2.x versions as well.

### DIFF
--- a/spec/test_doubles.rb
+++ b/spec/test_doubles.rb
@@ -118,6 +118,15 @@ module TestDoubles
     end
   end
 
+  class NamedArgumentOrganiser
+    extend LightService::Organizer
+
+    def self.call(number:)
+      with(number: number) # rubocop:disable Style/HashSyntax
+        .reduce([AddsTwoAction])
+    end
+  end
+
   class NotExplicitlyReturningContextOrganizer
     extend LightService::Organizer
 
--- a/spec/testing/context_factory_spec.rb
+++ b/spec/testing/context_factory_spec.rb
@@ -60,3 +60,18 @@ describe 'ContextFactory - used with AdditionOrganizer' do
     end
   end
 end
+
+describe 'ContextFactory - used with NamedArgumentOrganiser' do
+  let(:organizer) { TestDoubles::NamedArgumentOrganiser }
+
+  # it's relevant to test this as handling of named arguments changed between ruby 2.7 and 3.0
+  it 'pass named arguments to the organiser' do
+    ctx =
+      LightService::Testing::ContextFactory
+      .make_from(organizer)
+      .for(TestDoubles::AddsTwoAction)
+      .with(number: 2) # rubocop:disable Style/HashSyntax
+
+    expect(ctx[:number]).to eq(2)
+  end
+end
--- a/lib/light-service/testing/context_factory.rb
+++ b/lib/light-service/testing/context_factory.rb
@@ -34,6 +34,8 @@ module LightService
       end
       # rubocop:enable Style/ArgumentsForwarding
 
+      ruby2_keywords :with if RUBY_VERSION >= "2.7"
+
       def initialize(organizer)
         @organizer = organizer
       end


adomokos:
Thank you! I'll add Ruby 3 to the GH actions to make sure this fixes what you're describing.

adomokos:
@JeanSebTr - I looked into your PR today.

It seems `ruby2_keywords` was added to Ruby in 2.7. What if somebody (hopefully not 🤞 ) is using LS with Ruby <2.7? Could we add a guard clause for your statement?

Also, I'd love to see a unit test added for this change. Can you help me with that? 

Thank you so much!

adomokos:
I added Ruby 3.0.2 build in [this PR](https://github.com/adomokos/light-service/pull/225).
Your change might fix [this](https://github.com/adomokos/light-service/blob/bd5c93622d6d01eeb28c9cd1034076f4e16db999/lib/light-service/testing/context_factory.rb#L29).

JeanSebTr:
> It seems `ruby2_keywords` was added to Ruby in 2.7. What if somebody (hopefully not 🤞 ) is using LS with Ruby <2.7? Could we add a guard clause for your statement?

There is one: `ruby2_keywords :with if RUBY_VERSION >= "2.7"` as most ruby stuff, `ruby2_keywords` is a normal method and not a special keyword. I'm open to suggestions if it should take a different form than an if modifier.

I'll look into adding a test for it!

JeanSebTr:
> Your change might fix [this](https://github.com/adomokos/light-service/blob/bd5c93622d6d01eeb28c9cd1034076f4e16db999/lib/light-service/testing/context_factory.rb#L29).

looks like fixing that cop would require using `(...)` which would require dropping support for ruby < 2.7

JeanSebTr:
Hi @adomokos !

I've rebased on main and added a spec covering this. I didn't manage to run them locally tho

JeanSebTr:
Okay, managed to run specs locally, fixed the spec and made sure it fail on 3.0 without the fix.

pioneerskies:
Regarding the conversion about the backward compatibility, I'd like to leave this link https://www.ruby-lang.org/en/news/2019/12/12/separation-of-positional-and-keyword-arguments-in-ruby-3-0/ as a reference into the PR.

spec/test_doubles.rb
```diff
@@ -122,7 +122,7 @@ class NamedArgumentOrganiser
     extend LightService::Organizer
 
     def self.call(number:)
-      with(number: number).reduce([AddsTwoAction])
```
adomokos: Shouldn't we keep this to prove the functionality? Can you make a Rubocop exception for the next line here?

spec/testing/context_factory_spec.rb
```diff
@@ -70,7 +70,7 @@
       LightService::Testing::ContextFactory
       .make_from(organizer)
       .for(TestDoubles::AddsTwoAction)
-      .with(number: 2)
```
adomokos: Same here, I'd just "please" Rubocop to allow the rule violation on next line only.

adomokos (commented):
I'd keep the tests as is, and allow exceptions on the next line only.

spec/test_doubles.rb
pioneerskies: I fear I'm not getting the point here, sorry :)

Let's firstly say that obviously wouldn't be a problem to drop a `rubocop:disable*`. Just wanna be sure to share comprehension.

But I think that the answer to the first question should be: no. To prove the functionality the ruby interpreter >=3 has to be able to parse the script and more specifically [this line](https://github.com/adomokos/light-service/pull/224/files#diff-17a3795f56e12613500ff7d30051980f41fa9f2017a7b466298c45c412b254edR30) w/o dropping an `ArgumentError`.

The error is dropped ATM because in >=3 you can't receive keyword arguments in a method with a signature such as `method(*args)`.

To prove the fix suffice to pass a keyword argument. AFAIK a keyword argument is a keyword argument disregarding the syntax used to write it.

BTW

- the test fails with ArgumentError with both syntax w/o the `ruby2_keywords` fix
- the test succeds w/o ArgumentError with both syntax with the `ruby2_keywords` fix

JeanSebTr: Sorry all, I though I could wing this PR lazily through GH UI only...

Interestingly `:arg => value` is actually picked up as a kw argument:
image

However, that doesn't seem as evident as using directly the form `arg: value`. For clarity I'll use this form and add the rubocop rule.

JeanSebTr:
Ok, sorry about all that back and forth. I've applied @adomokos suggestions and ran both rspec and rubocop locally. Re-tested as well without the fix and the spec does fail.

spec/testing/context_factory_spec.rb
```diff
@@ -60,3 +60,18 @@
...
+describe 'ContextFactory - used with NamedArgumentOrganiser' do
+  let(:organizer) { TestDoubles::NamedArgumentOrganiser }
+
+  # it's relevant to test this as handling of named arguments changed between ruby 2.7 and 3.0
+  it 'pass named arguments to the organiser' do
+    ctx =
+      LightService::Testing::ContextFactory
+      .make_from(organizer)
+      .for(TestDoubles::AddsTwoAction)
+      .with(number: 2) # rubocop:disable Style/HashSyntax
```
adomokos, line 73: 😍

adomokos (approved):
Can I get one more reviewer on this PR? If not I'll merge it in 48 hours.
Thank you, @JeanSebTr for your patience and persistence with this change.

spec/test_doubles.rb
pioneerskies: Just saying: that’s not that interesting to me, because

`:foo => ‘bar’`

Is the really same thing as

`foo: ‘bar’`

Having in both cases Symbol as key and String as value.

This is AFAIK and for the sake of discussing 🙂

Thank you all for your time and effort on this 🚀

pioneerskies:
I already participated in the PR, deepened the matter, had the same need on a project and poor-man-tested the solution using 2.6 and 3.1: FWIW LGTM 🙂

JeanSebTr:
@adomokos  On our side, we're not in a hurry to migrate to ruby 3, so I don't mind if you prefer waiting for other stuff to release a version.

PR #232, opened 2022-03-30
Add reduce_case

kphan32:
Was looking for a when-else reducer but couldn't find one, so I made this.

`reduce_when` takes in:
- key to the value to perform the conditional logic on
- hash of values to array of steps
- named parameter `els` (else is reserved) that is an array of steps

Example
```
reduce_when(
  :incr_num,
  {
    :one => [TestDoubles::AddsOneAction],
    :two => [TestDoubles::AddsTwoAction],
    :three => [TestDoubles::AddsThreeAction]
  },
  :els => [TestDoubles::FailureAction]
)
```


adomokos:
@kphan32, thank you for the PR! Can you please provide a bit of context on where you think this `reduce_when` would be handy? I'd like to understand the use case a bit better.

adomokos:
Also, you might want to rebase from `main`, I bumped the Ruby versions last night.

kphan32:
WDYT of having the second parameter be named as `is`? That way it could be read as `reduce_when :key is ___ els ___`?

kphan32:
I took inspiration mainly from using Kotlin previously, which had a [`when` expression](https://kotlinlang.org/docs/control-flow.html#when-expression).

adomokos:
This helps. I like Kotlin's `when`, it's a decent-enough pattern matching attempt from JetBrains.

adomokos:
> WDYT of having the second parameter be named as is? That way it could be read as `reduce_when :key is ___ els ___`?

This could work. Can you show me what you mean by a working prototype?

pioneerskies:
Lovely :) 🙏🏻

Just my 2 cents about what I'd mainly expect from the interface:

```ruby
reduce_case(
  :key,
  when: {
    one: [TestDoubles::AddsOneAction],
    two: [TestDoubles::AddsTwoAction]
  },
  els: [TestDoubles::FailureAction]
)
```

because it would really quack like https://ruby-doc.org/core-3.1.1/doc/syntax/control_expressions_rdoc.html#label-case+Expression. Given the ruby expression is called `case`, I'd opt to name the reducer with the same word.

But I'll be obviuosly grateful for any decision <3

lib/light-service/organizer/reduce_when.rb
```diff
@@ -0,0 +1,19 @@
+module LightService
+  module Organizer
+    class ReduceWhen
+      extend ScopedReducable
+
+      def self.run(organizer, value_key, cases, else_steps)
+        lambda do |ctx|
+          return ctx if ctx.stop_processing?
+
+          steps = cases[ctx[value_key].to_sym] || else_steps
```
pioneerskies, line 10: As far as I understand, `ctx[value_key].to_sym` means that we support only symbols as possible values.

```ruby
TestReduceWhen.call(:number => 0, :incr_num => 2)
```

e.g., won't be supported. I'm imagining something like

```ruby
matched_case = cases.keys.find { |when| when.eql?(ctx[value_key]) }
steps = cases[matched_case] || else_steps
```

which should introduce support for a lot - if not all - value types, e.g:

```ruby
reduce_when(
          :current_value,
          {
            1 => [TestDoubles::AddsOneAction],
            [:ok, 'one'] => [TestDoubles::AddsTwoAction],
            :one => [TestDoubles::AddsThreeAction],
            FooObject => [TestDoubles::AddsFooAction]
          },
          :els => [TestDoubles::FailureAction]
        )
```

What do you think about the idea? Please consider I'm writing down the code w/o any sort of test/try.

I hope to not have misunderstood it all 🤞

pioneerskies: > n.d.r. maybe `when === ctx[value_key]` would be better that `.eql?` becouse it's the same comaprison method used by the `case` expression. But it hopefully it does not change the concept I was trying to deliver ;)

adomokos:
I do like what @pioneerskies suggested here:

```ruby
reduce_case(
  :key,
  when: {
    one: [TestDoubles::AddsOneAction],
    two: [TestDoubles::AddsTwoAction]
  },
  els: [TestDoubles::FailureAction]
)
```

I also like the idea of making the matching more flexible (goes against strongly typed languages, but we are talking Ruby here).

What do you think about making those changes @kphan32?

kphan32:
@adomokos Those sound good to me, will do :+1:

kphan32:
Though, `when` is a keyword so it'll end up being `wen`

adomokos:
> Though, when is a keyword so it'll end up being wen. Not sure how I feel about that, WDYT?

Yeah, `wen` is awkward.

What do you think about this?

```ruby
reduce_case(
  :key,
  :cases => {
     one: [TestDoubles::AddsOneAction],
     two: [TestDoubles::AddsTwoAction]
  },
  :els => [TestDoubles::FailureAction]
)
```

Let me know if @pioneerskies (or anybody else) has better ideas.

pioneerskies:
This is what I'd do in order to achive the most expressive interface bypassing the problem of having keywords

```ruby
module LightService
  module Organizer
    class ReduceWhen
      extend ScopedReducable

      class Arguments
        attr_reader :value, :when, :else

        def initialize(**args)
          validate_arguments(**args)
          @value = args[:value]
          @when = args[:when]
          @else = args[:else]
        end

        private

        def validate_arguments(**args)
          raise(
            ArgumentError,
            "Expected keyword arguments: [:value, :when, :else]. Given: #{args.keys}"
          ) unless args.keys.intersection(mandatory_arguments).count == mandatory_arguments.count
        end

        def mandatory_arguments
          %i[value when else]
        end
      end

      def self.run(organizer, **args)
        arguments = Arguments.new(**args)

        lambda do |ctx|
          return ctx if ctx.stop_processing?

          matched_case = arguments.when.keys.find { |k| k.eql?(ctx[arguments.value]) }
          steps = arguments.when[matched_case] || arguments.else

          ctx = scoped_reduce(organizer, ctx, steps)

          ctx
        end
      end
    end
  end
end
```

This way the example of use would be

```ruby
reduce_case(
  value: :foo,
  when: {
    'Foo' => [MyAction],
    bar: [MyOtherAction]
  },
  else: [AnotherAction]
)
```

Don't know if it fits, if it's worth the complexity, etc.

Hope it someway helps :)

kphan32:
@pioneerskies definitely a fan of this, and I don't mind the complexity.

adomokos:
This is looking good, I'll merge it. Can you, @kphan32 add README docs to it?

kphan32:
@adomokos Updated :+1:

adomokos (approved):
Looks good! Thank you, will merge.

adomokos:
Thank you @kphan32 for this great addition to LS! 💯

PR #194, opened 2020-06-03
Add Rails generators

gee-forr:
TL;DR - LightService is now bundled with generators for use within Rails.

As part of my drive to ensure LightService is adopted as much as possible with the people I work with, I thought providing rails generators would be a really useful way to help developers get traction with building software using LS.

This is a rough work-in-progress for now, but I thought it would be useful to get the code in early, and get any feedback, guidance, or help, before I go too far in a potentially wrong direction.

@adomokos - thanks again for this wonderful gift we rely on everyday. I hope you like the PR and find it a good fit for the project.

### Todo

- [x]  Update readme with generator documentation
- [x]  Ensure that generators work when LS is used as a gem (have only built them direct in rails right now, and they probably won't work)
- [X]  Generate actions
- [X]  Generate rspec specs for actions, unless `--no-tests`
- [x] Generated tests should all be set to "pending" with `xit`
- [x]  Add usage files for action generator
- [X]  Support module namespaced actions
- [X]  Support custom actions dir (defaults to `/app/actions`)
- [X]  Support defining expects and promises on the CLI
- [x]  Generate organizers
- [x]  Generate rspec specs for organizers, unless `--no-tests`
- [x]  Add usage files for organizer generator
- [x]  Support module namespaced organizers
- [x]  Support custom organizer dir (defaults to `/app/organizers`)
- [x]  DRY up code into modules
- [x]  Functional tests for generated actions
- [x]  Functional tests for generated organizers
- [ ]  ~~Generated tests for actions can execute via rspec~~ - not viable
- [ ]  ~~Generated tests for organizers can execute via rspec~~ - not viable
- [x] Address all rubocop violations

Issue #153, opened 2018-07-11
Hooks are invoked only on the first call


charlie-wasp:
Hello!

Today I stumbled upon some strange issue. It seems, that hooks (in my case `after_action`) are invoked only on the first `.call`.

Here is, how my method `call` looks like in the `Organizer` class:

```ruby
def self.call(chat_id:, user_to_ban_id:, voter:, vote: :for)
  with(
    chat_id: chat_id,
    user_to_ban_id: user_to_ban_id,
    voter: voter,
    vote: vote.to_sym,
    votes_storage: VotesStorage.new(user_to_ban_id),
  )
    .reduce(
      IncrementVotesCountAction,
      FetchVoteResultsAction,
      ResolveVotingAction,
    )
end
```

adomokos:
Thank you for reporting it!

I added [this spec](https://github.com/adomokos/light-service/commit/74189a83aa7c70bd19830c96c3f33004dfcfde2d) to verify this issue.  The `:number` in the result hash should be 1 (just like the one above it), but I expect 6 to get the test pass, as the callbacks are ignored. I'll have to rethink how the callbacks are kept for an organizer for subsequent calls.

The problem is that the callback is saved under the `data[:_after_actions]`, however, the second time the Organizer is invoked, it creates a new `LightService::Context` and the `:_after_actions` and `:_before_actions` callbacks are lost. :cry:

adomokos:
I started thinking about this. One way to solve this is to save the action specific callbacks in the current thread, something like this:

```ruby
Thread.current[:some_action_callbacks] = [proc1, proc2]
```

That way we could keep those callbacks around for multiple calls.

Thoughts?

charlie-wasp:
@adomokos sorry for not answering, been busy 😞  First I should sort out a whole architecture, and then maybe I would be able to suggest something :smile:

charlie-wasp:
Quick (and therefore maybe stupid) question: why we should nullify `@after_actions` in the first place?

klamontagne:
Hello, I was trying to use after_callbacks to add intrumentation to a light-service (via an included module) when I hit this bug. I was expecting to declare `after_actions` once in the class definition and have the instrumentation method run on every call (like `around_each` would help doing). The README example make it seem like this "static" implementation is intended. However, the spec uses it as a "dynamic" injection of the method (at runtime, before the call).

Maybe there should be a distinction between "static" callbacks and the current dynamic behavior.

[...]

PR #236, opened 2022-06-27
Relax activesupport dependency

alessandro-fazzi:
Given #226 and the bleeding announced in https://github.com/adomokos/light-service/issues/153#issuecomment-528679739 the goal of this PR is to relax the dependency this gem has on ActiveSupport.

**Why not removing it completely?**

It turns that the only part of the gem which continues to require ActiveSupport are the Rails generators. They were introduced recently in https://github.com/adomokos/light-service/pull/194. Thus I cannot afford to remove ActiveSupport with a one-shot PR. Nor I want to take decisions about how to continue from now on w/o including maintainer and community into the discussion.

**Is it worth to add more dependencies w/o removing ActiveSupport?**

Ideally not, but from a transitional perspective, yes. Now we have a single well scoped feature requiring ActiveSupport instead of having code bites spread all over the code base.
I'm bringing the gem in a more free and easier to manage position in regard of AS dep.

**What about the two new dependencies?**

I added https://dry-rb.org/gems/dry-inflector/ and https://github.com/schmidt/structured_warnings

Both were choosen because they are small, dependency-free gems with single functionality.

`dry-inflector` is from the (beloved by me) dry-rb ecosystem (https://dry-rb.org/). Neat, iper-atomic and well written. It is the Hanami inflector.

`structured_warnings` is a neat, very small gem; it has not been updated recently, but IMO it doesn't need to be. It works on  ruby 3.1 and its approach is simple and far (FAR) less polluting than ActiveSupport's one. I was able to implement a custom matcher for `rspec` too, writing expressive tests. Moreover its interface made possible to do a surgical substitution in the actual code base.

___

From my point of view, being the maintainer of the project, I'd think about splitting rails generators into a separate gem. This would enclose ActiveSupport and Rails-only things in cleaner context, easier to maintain. And the test suite of the "core" gem would lose all the complexity inherent to testing against all the possible AS versions.

I thought to open this one as a draft, but I decided to concretely propose something :) No problem turning it into draft or to use this just as a discussion desk and then close it.

Thanks for your attention.

### DIFF
--- a/lib/light-service.rb
+++ b/lib/light-service.rb
@@ -1,5 +1,5 @@
 require 'logger'
-require 'active_support/core_ext/string'
+require 'i18n'
 
 require 'light-service/version'
 
--- a/lib/light-service.rb
+++ b/lib/light-service.rb
@@ -1,5 +1,6 @@
 require 'logger'
 require 'i18n'
+require 'structured_warnings'
 
 require 'light-service/version'
 
--- a/lib/light-service/action.rb
+++ b/lib/light-service/action.rb
@@ -1,4 +1,4 @@
-require 'active_support/deprecation'
+require 'structured_warnings'
 
 module LightService
   module Action
@@ -9,7 +9,7 @@ module LightService
     def self.included(base_class)
       msg = "including LightService::Action is deprecated. " \
             "Please use `extend LightService::Action` instead"
-      ActiveSupport::Deprecation.warn(msg)
+      warn(StructuredWarnings::DeprecatedMethodWarning, msg)
       base_class.extend Macros
     end
 
--- a/light-service.gemspec
+++ b/light-service.gemspec
@@ -18,6 +18,8 @@ Gem::Specification.new do |gem|
   gem.required_ruby_version = '>= 2.6.0'
 
   gem.add_runtime_dependency("activesupport", ">= 4.0.0")
+  gem.add_runtime_dependency("i18n")
+  gem.add_runtime_dependency("dry-inflector")
 
   gem.add_development_dependency("generator_spec", "~> 0.9.4")
   gem.add_development_dependency("test-unit", "~> 3.0") # Needed for generator specs.
--- a/light-service.gemspec
+++ b/light-service.gemspec
@@ -20,6 +20,7 @@ Gem::Specification.new do |gem|
   gem.add_runtime_dependency("activesupport", ">= 4.0.0")
   gem.add_runtime_dependency("i18n")
   gem.add_runtime_dependency("dry-inflector")
+  gem.add_runtime_dependency("structured_warnings")
 
   gem.add_development_dependency("generator_spec", "~> 0.9.4")
   gem.add_development_dependency("test-unit", "~> 3.0") # Needed for generator specs.
--- a/spec/acceptance/include_warning_spec.rb
+++ b/spec/acceptance/include_warning_spec.rb
@@ -1,16 +1,17 @@
 require 'spec_helper'
+require 'structured_warnings_helper'
 
 describe "Including is discouraged" do
   context "when including LightService::Organizer" do
     it "gives warning" do
       expected_msg = "including LightService::Organizer is deprecated. " \
                      "Please use `extend LightService::Organizer` instead"
-      expect(ActiveSupport::Deprecation).to receive(:warn)
-        .with(expected_msg)
 
-      class OrganizerIncludingLS
-        include LightService::Organizer
-      end
+      expect do
+        class OrganizerIncludingLS
+          include LightService::Organizer
+        end
+      end.to warn_with(StructuredWarnings::DeprecatedMethodWarning, expected_msg)
     end
   end
 
@@ -18,12 +19,12 @@ describe "Including is discouraged" do
     it "gives warning" do
       expected_msg = "including LightService::Action is deprecated. " \
                      "Please use `extend LightService::Action` instead"
-      expect(ActiveSupport::Deprecation).to receive(:warn)
-        .with(expected_msg)
 
-      class ActionIncludingLS
-        include LightService::Action
-      end
+      expect do
+        class ActionIncludingLS
+          include LightService::Action
+        end
+      end.to warn_with(StructuredWarnings::DeprecatedMethodWarning, expected_msg)
     end
   end
 end
--- a/lib/generators/light_service/generator_utils.rb
+++ b/lib/generators/light_service/generator_utils.rb
@@ -26,6 +26,10 @@ module LightService
       end
 
       def create_required_gen_vals_from(name)
+        # Not using dry-inflector here, because generators are used only
+        # within Rails project thus we can relay on ActiveSupport presence.
+        # Maybe plan to split LightService::Generators into a separate
+        # gem (e.g.: light_service-rails-generators)
         path_parts = name.underscore.split('/')
 
         {
--- a/lib/light-service/context.rb
+++ b/lib/light-service/context.rb
@@ -1,4 +1,4 @@
-require 'active_support/deprecation'
+require 'structured_warnings'
 
 module LightService
   module Outcomes
@@ -61,7 +61,7 @@ module LightService
     def outcome
       msg = '`Context#outcome` attribute reader is ' \
             'DEPRECATED and will be removed'
-      ActiveSupport::Deprecation.warn(msg)
+      warn(StructuredWarnings::DeprecatedMethodWarning, msg)
       @outcome
     end
 
@@ -103,7 +103,7 @@ module LightService
     def skip_all!(message = nil)
       warning_msg = "Using skip_all! has been deprecated, " \
                     "please use `skip_remaining!` instead."
-      ActiveSupport::Deprecation.warn(warning_msg)
+      warn(StructuredWarnings::DeprecatedMethodWarning, warning_msg)
 
       skip_remaining!(message)
     end
--- a/lib/light-service/localization_adapter.rb
+++ b/lib/light-service/localization_adapter.rb
@@ -1,5 +1,13 @@
+require "dry/inflector"
+
 module LightService
   class LocalizationAdapter
+    attr_reader :inflector
+
+    def initialize
+      @inflector = Dry::Inflector.new
+    end
+
     def failure(message_or_key, action_class, i18n_options = {})
       find_translated_message(message_or_key,
                               action_class,
@@ -38,7 +46,7 @@ module LightService
     end
 
     def i18n_scope_from_class(action_class, type)
-      "#{action_class.name.underscore}.light_service.#{type.to_s.pluralize}"
+      "#{inflector.underscore(action_class.name)}.light_service.#{inflector.pluralize(type.to_s)}"
     end
   end
 end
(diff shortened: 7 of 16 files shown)


adomokos:
This is a great PR! I thank you for the work you put into it, I am going to merge it.
If I understand this correctly, after your change, the only dependency we have for AS is the Rails generator.

You wrote:

> From my point of view, being the maintainer of the project, I'd think about splitting rails generators into a separate gem

I am all for doing this, but we should move LightService under an org, and have the gem, and this other one with Rails generator should be there. What do you think?

light-service.gemspec
```diff
@@ -18,6 +18,9 @@ Gem::Specification.new do |gem|
   gem.required_ruby_version = '>= 2.6.0'
 
   gem.add_runtime_dependency("activesupport", ">= 4.0.0")
+  gem.add_runtime_dependency("i18n")
+  gem.add_runtime_dependency("dry-inflector")
+  gem.add_runtime_dependency("structured_warnings")
```
adomokos, line 23: Oh, I missed these additions. So instead of having just AS, we are adding 3 more dependencies.
I wish we could achieve the same functionality with no additional gems. But maybe I am just asking too much.

I wonder what ppl think about this change.

pioneerskies:
> This is a great PR! I thank you for the work you put into it, I am going to merge it. If I understand this correctly, after your change, the only dependency we have for AS is the Rails generator.
> 
> You wrote:
> 
> > From my point of view, being the maintainer of the project, I'd think about splitting rails generators into a separate gem
> 
> I am all for doing this, but we should move LightService under an org, and have the gem, and this other one with Rails generator should be there. What do you think?

Reading again the code, I’d like to do a couple of little refinements. But I’ll let the discussion go further before to do anything else.

I think, consolidated the choice to split the code, the organization would be the more elegant solution and the better guarantee for the community.
But pragmatically speaking it wouldn’t be mandatory.

I don’t think an organization would be that big management complication, but it’s a personal point of view and a matter to think about before going all-in.

light-service.gemspec
pioneerskies: We’re adding only 2 more dependencies, since i18n is already in through AS at the moment.

I explicitly required it as a dep in order to have a functioning gemspec even considering AS removal and without leaving remainders for future work.

Speaking about the balance between AS and the two new gems: my sentiment is that with the two gems we bring less code than with AS alone. We should analyze it more scientifically tho.

For sure we have the achievement to not have a big rails ecosystem’s gem, leaving LS more desirable as cross-framework and franework-less solution.

Speaking of deprecation warnings: I previously tried to implement them using Gem::Deprecate, but it resulted to me as an inadequate tool out of its absolute standard concept. Hard to adapt. I think it would be feasible to write a dedicated Logger and implement some logic inside it in order to achieve a simple custom deprecation warner. This would be intended as a pragmatic and simpler alternative to structured_warning gem.
If you’d like to check a draft implementation I could try to write a separate draft PR in the second half of July.

About the inflector I’m really opinionated: I would not even try to reinvent the wheel on such a matter. Moreover I sincerely  invite you to read dry-inflector’s source code: it’s astonishing small, tidy and simple; and it’s a completely standalone gem. Obviously we could copy that few methods we need, but pro and cons considered I think it’s a solid choice to bring it in.

Thanks for the feedback 😊

lib/light-service.rb
```diff
@@ -1,5 +1,6 @@
 require 'logger'
-require 'active_support/core_ext/string'
+require 'i18n'
+require 'structured_warnings'
```
gee-forr, line 3: If we're requiring `structured_warnings` in `light-service.rb` we probably don't need to re-require them again in the rest of the files, right?

spec/acceptance/include_warning_spec.rb
```diff
@@ -1,29 +1,30 @@
...
-        .with(expected_msg)
 
-      class OrganizerIncludingLS
-        include LightService::Organizer
-      end
+      expect do
+        class OrganizerIncludingLS
+          include LightService::Organizer
+        end
+      end.to warn_with(StructuredWarnings::DeprecatedMethodWarning, expected_msg)
```
gee-forr, line 14: These tests are great! We should remember to remove them when the actual removal takes place. 👍🏻

gee-forr (commented):
Looking good... one note about the re-requiring of `structured_warnings` but everything else looks good.

gee-forr:
@adomokos - if you're happy to move forward with this, I am. I think adding two more dependencies is a small and temporary price to pay for what it unlocks for the future.

If you want to move forward with the org + separate rails gem, then, @pioneerskies could I ask that you also add deprecation warnings to the rails generators?

lib/light-service.rb
pioneerskies: Totaly agree: OR in the manifest OR in single files. In both places is just confusing and conceptually wrong.

spec/acceptance/include_warning_spec.rb
pioneerskies: Given this discussion https://github.com/adomokos/light-service/pull/236#discussion_r908968696 it's still to take a decision about the use of that library.

That said and considering that whatever implementation will be it will have a test: remembering to remove deprecation from code will bring a red CI. That should be a good post-it to remove the test too ;)

lib/light-service/action.rb
```diff
@@ -9,7 +9,7 @@ def self.extended(base_class)
     def self.included(base_class)
       msg = "including LightService::Action is deprecated. " \
             "Please use `extend LightService::Action` instead"
-      ActiveSupport::Deprecation.warn(msg)
+      warn(StructuredWarnings::DeprecatedMethodWarning, msg)
```
pioneerskies, line 12: Here and wherever we thrown a `warn`, if this gem will remain in, I'd like to have a `class LightService::Deprecation < StructuredWarnings::Base; end` to use insted of directly expose the gem's built-in.

gee-forr: I'm completely for that course of action.

adomokos:
I am afk, but will pick this up, thoroughly review it, and merge it next week. Thanks for your patience there!

pioneerskies:
> I am afk, but will pick this up, thoroughly review it, and merge it next week. Thanks for your patience there!

I'll be on vacation starting from 1st until 19th July. In the mean time feel free to do what it's needed with "my" code. I've already annoted and subscribed desirable updates and expressed all the opinions I had to. So feel free to act w/o my intervention in the coming weeks :)

Thank you all so much.

adomokos:
Hello @pioneerskies! Sorry for keeping this PR open for so long.

I thought about this while I was away. How hard would it be not to introduce two (albeit small) gems to light-service, but leverage pure Ruby to accomplish this functionality? I'd love to have LS be outside gem-dependent-free if possible in the future.

pioneerskies:
> Hello @pioneerskies! Sorry for keeping this PR open for so long.
> 
> I thought about this while I was away. How hard would it be not to introduce two (albeit small) gems to light-service, but leverage pure Ruby to accomplish this functionality? I'd love to have LS be outside gem-dependent-free if possible in the future.

Hi there,

AFAIK deprecations could be simplified down to a custom `Logger`: if my assumption would be right, then writing 1st party code would be easy. I mean "easy" as "not complex", not as "fast", but that's it.

The inflector instead would be different: `dry-inflector` is the smallest, most low-complexity thing I've found around. Given we don't want to use any 3rd party dep, then I'd opt for brutally copy/paste-ing the gem's `lib/` into our codebase. That's why I won't be able to design something different and more simple than that code and reinventing the wheel would be wasted time with a worse final result (since I'm honestly light years far from Luca's and Peter's design/coding skills :) )

Conclusion: a custom "deprecation logger" + a savvy copy-paste should result in a quite small and not complex goal to achieve.

adomokos:
I see adding 3 ruby gems with this PR:

1. `i18n` - which needs `concurrent-ruby`
2. `dry-inflector` - no external library dependency
3. `structured_warnings` - no external library dependency

I do like the intent of this PR, but I am now hesitant to merge it. We are adding quite a bit of dependency, and I don't see the path of cutting out `activesupport` altogether. I wonder if we should keep this PR open, create a new gem `light-service-rails`, extract the Rails generator dependencies into that, and then merge this PR. That way we have a guaranteed path of reducing the external gem dependency.

What do you think about that?

I am happy to create a new repo for `light-service-rails`, I am not sure if I'd create a new organization for it yet. I'd be happy to add you as maintainer to this new project if you're willing to help me extract the Rails and AS dependencies. What do you think?

pioneerskies:
I like to re-underline that `i18n` here is not added, but just made explicit: it's already an [AS's dependency](https://github.com/rails/rails/blob/main/activesupport/activesupport.gemspec). As it stands it's just a peer dependence through AS but it is explicitly used for failure message translation. Having it as a direct dependence was a way to underline the fact that even in the will to remove AS we'll have to retain `i18n`.

IMO we should be able to avoid `i18n` dependency too, making it opt-in: if an implementor would like to use translated messages she should add `i18n` to her bundle and LS would pick it up automatically. If absent and a `Symbol` is passed as failure message, LS could drop a warning ad just `.to_s` the symbol.

I know this i18n-related discourse may seem OT, but I consider part of the "dependent-free" path you are picturing.

Coming on the spot of your comment now :)

> I do like the intent of this PR, but I am now hesitant to merge it

I'm not only okay with that, but I'm really happy with that. My intention was to open a discussion bringing a practical substrate to it.

> We are adding quite a bit of dependency

I think we're trading dependencies here, more than adding. The biggest fault of this PR, in my opinion, is that it moves from one dep from another w/o layering them with first party _wrappers_. Wrappers would be a refactoring pattern making possible to plug in/out external deps or first party code w/o touching other parts of code. E.g.:

```ruby
# lib/light-service/inflector.rb

require "dry/inflector"

module LightService
  class Inflector
    attr_reader :inflector

    def initialize
      @inflector = Dry::Inflector.new
    end

    def unerscore(string)
      inflector.underscore(string)
    end

    def pluralize(string)
      inflector.underscore(string)
    end
  end
end
```

```ruby
# lib/light-service/localization_adapter.rb

module LightService
  class LocalizationAdapter
    attr_reader :inflector

    def initialize
      @inflector = LightService::Inflector
    end

    # [...]

    def i18n_scope_from_class(action_class, type)
      "#{inflector.underscore(action_class.name)}.light_service.#{inflector.pluralize(type.to_s)}"
    end
  end
end
```

This way we'd be able to swap in/out different external deps or get rid of them and write our own or anything we'd like to, with a single point of responsibility

> I wonder if we should keep this PR open, create a new gem light-service-rails, extract the Rails generator dependencies into that, and then merge this PR

That path would be for sure the more profitable one.

Current PR had just the goal to make things move forward given the actual code base and to poll the interest around the proposed mid-term goal. 

Keep this one open as draft and we'll understand if it will be better to update it or to close it for a new one.

> I am happy to create a new repo for light-service-rails, I am not sure if I'd create a new organization for it yet. I'd be happy to add you as maintainer to this new project if you're willing to help me extract the Rails and AS dependencies. What do you think?

- don't bother to go for the org atm: let's do some work and then you'll think about organizational matters
- I'll be happy to help until the end of September; then my baby-girl will come to the world and I'm ready to bet I'll need to take a pause from free-time work for a while :)
- let's check for @gee-forr availability to help too; even just as reviewer. As original author of generator's code and as active user of the feature he may be a great booster

adomokos:
If and when we go with the route of a separate gem version for `light-service-rails`, we will need to bump the major version to 1.

adomokos:
Closing this, as AS was removed with [this PR](https://github.com/adomokos/light-service/pull/246).

PR #238, opened 2022-12-05
Implement a built-in, dependency free localization adapter

pioneerskies:
## Preface

This proposal takes step from the discussion in https://github.com/adomokos/light-service/pull/236.

To the attention of both @adomokos and @gee-forr 

The overall plan in my head, at the moment, is the following

1. decouple light-service from I18n (**this PR**)
  1.1. preserve a basic translation functionality
  1.2 do not break compatibility with existing implementations relying on I18n
2. strip out `LightService::I18n::LocalizationAdapter` into its own `light_service-i18n` gem
  2.1 require gem installation to retain full compatibility with existing implementations relying on I18n
3. decouple light-service from `ActiveSupport::Deprecation`
  3.1 preserve full functionality through the use of a simple custom logger built on top of `Logger`
4. decouple light-service from `ActiveSupport::Inflector`
  4.1 write a `LightService::Inflector` class responding only to the needed inflection methods (probably `#underscore` and `#pluralize`)
  4.2 `LightService::Inflector` will delegate its capabilities to `dry-inflector` (which is a zero-deps gem)
5. strip out `LightService::Generators` into its own `light_service-rails` gem
6. release light-service 1.0
  6.1 shut out upgrade instructions in changelog, requiring installation of new detached gems in order to retain moved functionalities (in my head this should avoid a long lasting deprecation path...but we'll have to find a place to discuss these sort of topics)
  6.2 think about renaming the gem to `light_service`. This will require a well thought strategy, if desirable at all.

Pictured this out, let's talk about current PR.

## Details

The built-in adapter is trivial and zero-deps.

This PR is thought to be the first step of a two step update concerning translations.

### First step (Current PR)

The gem ships a localization adapter that does not rely on i18n gem. It has not feature parity because it can't interpolate variables.

The gem, when loaded, will check for `I18n` presence into the global space. If it's not present, then it will use the built-in adapter. If it is present, then an alternative `LightService::I18n::LocalizationAdapter` will be automatically used.

`LightService::I18n::LocalizationAdapter` is exactly the same localization adapter shipped previously.

The README is updated to reflect the new scenario.

Keep in mind that, as it stands, i18n gem is required indirectly by light-service as an activesupport dependency, thus it will be automatically used for translations. This means that this PR won't affect any existing implementation.

### Step two (future PR)

`LightService::I18n::LocalizationAdapter` and its specs will be moved into a new `light_service-i18n` gem (or `light_service-translation`).

light-service will stop to automatically choose the translation adapter, always defaulting to the built-in one.  light_service-i18n will instead forcibly configure itself as translation adapter when installed.

### About branches

Alternatively to request to merge on `main`, a new development branch could be created on the original repository and the PRs could be pointed to that branch. Major version bumps could this way be delayed to the future when generators will be migrated to dedicated gem and the current inflector (activesupport) will be changed to something else. This development branch should have a name that represents the overall roadmap of removing activesupport/rails as dependency


gee-forr:
hey @pioneerskies - this is a very very comprehensive PR which I'll go through in more detail a little later... I did want to immediately note one point I was thinking on re your notes above:

I'm unclear if the localisation gem will be required once it's completely implemented, or if it will remain optional forever?

Also - I'm totally onboard with moving the rails specific code to a separate gem. It will allow it to evolve and take on more functionality which arguably happens at a quicker pace due to the generators missing functionality for other testing libs.

gee-forr:
Sorry - one last thing, which is off-topic, but I'd like to build some **optional** smarter validation into the `expects` and `promises` macros, maybe around types or expected interfaces vs. the simple validation right now. And once that's done, we can call it a 1.0 release. Obv. @adomokos would need to be on board with this of course. I know he wants to keep things simple and light -- hence the name :)

pioneerskies:
> I'm unclear if the localisation gem will be required once it's completely implemented, or if it will remain optional forever?

@gee-forr In my vision it will remain optional; that's why I've thought about implementing a built-in, very simple translation mechanism: offer a sufficient solution without involving further dependencies. That gem would require `i18n` without version constraint as a dependency.

All my vision is a proposal, a reasoning and completely open to discussion.

pioneerskies:
> Sorry - one last thing, which is off-topic, but I'd like to build some **optional** smarter validation into the `expects` and `promises` macros, maybe around types or expected interfaces vs. the simple validation right now. And once that's done, we can call it a 1.0 release. Obv. @adomokos would need to be on board with this of course. I know he wants to keep things simple and light -- hence the name :)

@gee-forr you might be interested into taking a look at https://github.com/pioneerskies/light_service-validated_context/

Even just to tinkle your ideas reading someone else's code.

adomokos:
Hey guys, @pioneerskies and @gee-forr!
I am going to set up an organization for `light-service`, move the project there, and add you both as a core contributors. That way we can add all "sidecar" gems there and can be maintained together.

We should do this in preparation for release v1.0. What do you think?

pioneerskies:
> Hey guys, @pioneerskies and @gee-forr! I am going to set up an organization for `light-service`, move the project there, and add you both as a core contributors. That way we can add all "sidecar" gems there and can be maintained together.
> 
> We should do this in preparation for release v1.0. What do you think?

I'm happy to partecipate with what I can bring, and moreover thankful of the proposal. And it seems a good fit given most recent discussed intentions.

What about the battle '-' vs. '_' in the naming? Better to think about it earlier or not a problem to do it later?

adomokos:
I created the org `light-service`, and added you two as members.
I'd like to reserve the right to create repos there, the first one I created is [light-service-rails](https://github.com/light-service/light-service-rails).

What do you guys think about `_` vs`-`? I can rename it if you want, but `light-service` has a ton of Google history already, Plus, I wonder if this rename would require a new RubyGems repo.

I'd like to keep `light-service` under my account until the 1.0 release, to see how this partnership works out. If all is well, I am happy to transfer the code under the `light-service` org. Right now, I'd like to use the organization as a repository for all "satellite" or "sidecar" Ruby gems we extract from light-service.

I'll also go with your suggestion, I'll create a new branch `v1`, and all this work you are doing should be merged into that.

adomokos:
I agree, this is the time to get the [name right](https://guides.rubygems.org/name-your-gem/). With the v1 release, we should fix it. I might have to delete the org and create a new one with proper naming.

pioneerskies:
> I created the org `light-service`, and added you two as members.

Yay

> I'd like to reserve the right to create repos there, the first one I created is [light-service-rails](https://github.com/light-service/light-service-rails).
 Uber-yay

> What do you guys think about `_` vs`-`? I can rename it if you want, but `light-service` has a ton of Google history already, Plus, I wonder if this rename would require a new RubyGems repo.

There are some technical implications in using `_` vs `-` in the name; the first I noticed is the standard naming used by bundler when creating a gem, the second is the naming convention enforced by `zeitwerk` which is probably something we'll like to face up given it's quite becoming a non-written standard (maybe it's just a rails "fault"? IDK, but I tend to do not think so).

I'd need to deepen this matter a bit before to spit out something opinionated :)

> I'd like to keep `light-service` under my account until the 1.0 release, to see how this partnership works out. If all is well, I am happy to transfer the code under the `light-service` org. Right now, I'd like to use the organization as a repository for all "satellite" or "sidecar" Ruby gems we extract from light-service.

100%. No more nor less

> I'll also go with your suggestion, I'll create a new branch `v1`, and all this work you are doing should be merged into that.

Sounds good to me. Thanks

adomokos:
K, both of you are members in the `light-service` organization. I couldn't use underscore there, as GH wouldn't allow it.
I created two empty [repos](https://github.com/orgs/light-service/repositories):

1. light_service-rails
1. light_service-i18n

I preserved the dash for `-rails` and `-i18n`. Holler (or better yet, modify) if you disagree. Let's use those repos as a future home of everything we extract from LightService.

I also created the `v1` branch, we should use that for the `light_service v1` release. LS should be lean and mean, have all the core logic, with (almost) zero external gem dependencies.

Thank you for all your contribution!

adomokos:
I wonder if you want to merge into `main` and not `v1`. This change looks good to me.

adomokos (commented):
Tiny documentation suggestion.

pioneerskies:
> I wonder if you want to merge into `main` and not `v1`. This change looks good to me.

Have to mumble about it until my next window, but in the mean time I'm happy to have you confirming that mine was an unobtrusive approach

adomokos:
Let's merge it into Main. This is a great change, thank you for it!

gee-forr:
> @gee-forr you may be interested into taking a look at https://github.com/pioneerskies/light_service-validated_context/
> Even just to tinkle your ideas reading someone else's code.

This is amazing... I think the interface is a little heavy, but we could definitely build off of it to make it "light"er.

gee-forr:
> Hey guys, @pioneerskies and @gee-forr! I am going to set up an organization for `light-service`, move the project there, and add you both as a core contributors. That way we can add all "sidecar" gems there and can be maintained together.

Awesome - I'm on board :) I'll start work on building the standalone rails gem during my December downtime.

pioneerskies:
@adomokos `v1` branch should be rebased on top of `main`. I do not open a PR because you'd have the same amount of work to do ^^"

TIA

adomokos:
Done.

PR #239, opened 2022-12-08
Remove i18n localization adapter

alessandro-fazzi:
## Preface

This PR is part of the plan discussed in #236 

- decouple light-service from I18n #238
  - preserve a basic translation functionality #238
  - do not break compatibility with existing implementations relying on I18n #238
- strip out LightService::I18n::LocalizationAdapter into its own light_service-i18n gem #239
  - require gem installation to retain full compatibility with existing implementations relying on I18n #239
- decouple light-service from ActiveSupport::Deprecation #241 
  - preserve full functionality through the use of a simple custom logger built on top of `Logger`
- decouple light-service from ActiveSupport::Inflector 
  - write a LightService::Inflector class responding only to the needed inflection methods (probably #underscore and #pluralize)
  - LightService::Inflector will delegate its capabilities to dry-inflector (which is a zero-deps gem)
- strip out LightService::Generators into its own light_service-rails gem
- release light-service 1.0
  - shut out upgrade instructions in changelog, requiring installation of new detached gems in order to retain moved functionalities (in my head this should avoid a long lasting deprecation path...but we'll have to find a place to discuss these sort of topics)
  - think about renaming the gem to light_service. This will require a well thought strategy, if desirable at all.

## Details

- removed `LightService::I18n::LocalizationAdapter`
- removed relative specs
- removed specific I18n configuration from spec_helper
- `LightService::Configuration.localization_adapter` no more gets automatically updated based on `i18n` gem presence in bundle
- `README.md` updated accordingly


pioneerskies:
@adomokos @gee-forr you're welcome to review.

I'm in doubt if README's documentation should be only a link to the sidecar gem's repo. I can see pros and cons. Opinions more than welcome - as always.

PR #241, opened 2022-12-10
Swap ActiveSupport::Deprecation with built-in LightSupport::Deprecation

pioneerskies:
## Preface

This PR is part of the plan discussed in #236 

- [x] decouple light-service from I18n #238
  - [x] preserve a basic translation functionality #238
  - [x] do not break compatibility with existing implementations relying on I18n #238
- [x] strip out LightService::I18n::LocalizationAdapter into its own light_service-i18n gem #239
  - [x] require gem installation to retain full compatibility with existing implementations relying on I18n #239
- [x] decouple light-service from ActiveSupport::Deprecation #241
  - [x] preserve full functionality through the use of a simple custom logger built on top of `Logger` #241
- [ ] decouple light-service from ActiveSupport::Inflector
  - [ ] write a LightService::Inflector class responding only to the needed inflection methods (probably #underscore and #pluralize)
  - [ ] LightService::Inflector will delegate its capabilities to dry-inflector (which is a zero-deps gem)
- [ ] strip out LightService::Generators into its own light_service-rails gem
- [ ] release light-service 1.0
  - [ ] shut out upgrade instructions in changelog, requiring installation of new detached gems in order to retain moved functionalities (in my head this should avoid a long lasting deprecation path...but we'll have to find a place to discuss these sort of topics)
  - [ ] think about renaming the gem to light_service. This will require a well thought strategy, if desirable at all.

## Details

Swapping from `ActiveSupport::Deprecation` to `LightService::Deprecation`.

`LightService::Deprecation` is a new built-in class simply built on top of `Logger`.

The only public method is `.warn` - it mimics AS interface.

The logger will log warnings to STDERR with the following format

```
[DEPRECATION] /Users/fuzzy/dev/light-service/lib/light-service/context.rb:106:in `skip_all!': Using skip_all! has been deprecated, please use `skip_remaining!` instead
```


adomokos (approved):
🚢

PR #243, opened 2023-02-27
Add Ruby 3.2 and ActiveSupport 7 to CI. Update checkout action version.

petergoldstein:
Also updated the pry restriction to enable Ruby 3.2 to run successfully.  Runs green on my fork.


.github/workflows/project-build.yml
```diff
@@ -13,16 +13,21 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu, macos]
-        ruby: [2.7.4, 3.0.2, 3.1.0]
-        gemfile: [activesupport_5, activesupport_6]
+        ruby: [2.7, '3.0', '3.1', '3.2']
+        gemfile: [activesupport_5, activesupport_6, activesupport_7]
+        exclude:
+          - ruby: '3.2'
+            gemfile: activesupport_5
```
adomokos, line 20: Why is excluded?

.github/workflows/project-build.yml
```diff
@@ -13,16 +13,21 @@ jobs:
...
     env: # $BUNDLE_GEMFILE must be set at the job level, so it is set for all steps
       BUNDLE_GEMFILE: gemfiles/${{ matrix.gemfile }}.gemfile
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
       - uses: ruby/setup-ruby@v1
         with:
           ruby-version: ${{ matrix.ruby }}
-      - run: bundle install
+          bundler-cache: true
```
adomokos, line 31: Do we need to run `bundle install`?

adomokos (commented):
Thank you for your contribution, I'll merge this. 🏅  I only had a couple of questions.

.github/workflows/project-build.yml
petergoldstein: @adomokos Running an explicit bundle install isn't required when using the `setup-ruby` task with `bundler-cache` set to true.

.github/workflows/project-build.yml
petergoldstein: Rails 5.x isn't supposed to be run with Ruby 3.2.  It may run successfully, or it may not, but it's an unsupported configuration.

adomokos (approved):
Thank you! 🚢

Issue #258, opened 2024-10-26
ActiveSupport removal did not work


adomokos:
`singularized` is still used [here](https://github.com/adomokos/light-service/blob/aa27ba194c3b46a052c270ed1dced397f2bc07d0/lib/light-service/organizer/iterate.rb#L11), and the spec is still passing. 😞

PR #259, opened 2024-10-26
Add back activesupport

### LINKED ISSUE #257: Not providing `expects` or `promises` values will error out
This is the error:

```
light-service-0.19.0/lib/light-service/context.rb:120:in `define_accessor_methods_for_keys':
undefined method `blank?' for an instance of Array (NoMethodError)
```

[Here](https://github.com/adomokos/light-service/blob/aa27ba194c3b46a052c270ed1dced397f2bc07d0/lib/light-service/context.rb#L119).

adomokos:
We are not ready to remove it. There are instances of `blank?` and `singularize` that are used in the code.

Removing `activesupport` still not fails any of the tests, as we have `generator_spec` that uses and installs `activesupport`.

Fixes #257  and #258.

### DIFF
--- a/lib/light-service/context.rb
+++ b/lib/light-service/context.rb
@@ -116,7 +116,7 @@ module LightService
     end
 
     def define_accessor_methods_for_keys(keys)
-      return if keys.blank?
+      return if keys.empty?
 
       Array(keys).each do |key|
         next if respond_to?(key.to_sym)
--- a/README.md
+++ b/README.md
@@ -9,8 +9,6 @@
 
 LightService is a powerful and flexible service skeleton framework with an emphasis on simplicity
 
-🔥 **It now comes with no external gem dependency.** 🔥
-
 ## Table of Contents
 - [Table of Contents](#table-of-contents)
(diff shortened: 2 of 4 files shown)


adomokos:
Addresses #260

wqsaali:
@adomokos also coding is using `underscore` method coming from `active_support/core_ext/string/inflections`

lib/light-service/context.rb
```diff
@@ -116,7 +116,7 @@ def stop_processing?
     end
 
     def define_accessor_methods_for_keys(keys)
-      return if keys.blank?
+      return if keys.empty?
```
wqsaali, line 119: @adomokos since we are adding back `activesupport`, do we really need to make this change? Instead, we could simply include `active_support/core_ext/object/blank` and use the same method without introducing additional changes.

adomokos: The goal is to yank Active Support dependency, entirely, for pure Ruby projects. Not sure how easily feasible (or how much needed) that is, though.

PR #274, opened 2025-06-30
Update README.md to reflect before and after actions do not execute on every call to an organizer.

RCulpepp-frankly:
Even though based on a previous bug report it seems as if this is not the intended behavior, it hasn't been addressed and the docs should reflect the current state not to confuse others.  I spent a little while debugging an issue around this and it'd be good to spare others that same frustration until a fix can be implemented.


adomokos (approved):
🚢

Issue #275, opened 2025-07-04
skip_remaining only applying to actions in reduce_if blocks


AaronCDee:
Hi, there.

I've picked up some strange behaviour with `skip_remaining!` in `reduce_if` blocks.
When `skip_remaining!` is called inside of a `reduce_if` set of actions, only the actions in the `reduce_if` block are skipped.

EG:
```ruby
def self.actions
  [
    reduce_if(->(ctx) { ctx.should_execute_conditional_actions }, [
      ConditionallyRunAction1, # NOTE: ctx.skip_remaining! called here
      ConditionallyRunAction2, # NOTE: This action is skipped as expected
    ]
    AlwaysRunAction, # NOTE: This action is NOT skipped
  ]
end
```

I'm not sure if this is a bug or if this is intended behaviour, but thought I would ask :).
Thanks, Aaron

adomokos:
That's how it was designed: always jumping out of the closest construct - iirc. We might need something like `skip_all_remaining!` to halt execution without errors.

PR #279, opened 2026-01-02
Add Skip All functionality

sncalvo:
# Description

This PR adds skip all functionality described [here](https://github.com/adomokos/light-service/issues/275).

While `context.skip_remaining!` skips actions in the current scope (e.g. inside a `reduce_if` or `iterate` block), sometimes you want to halt the entire execution of the organizer and any parent organizers. You can do this by calling `context.skip_all_remaining!`.

Consider this example:

```ruby
def self.actions
  [
    reduce_if(->(ctx) { ctx.should_execute_conditional_actions }, [
      ConditionallyRunAction1, # calls ctx.skip_all_remaining!
      ConditionallyRunAction2, # skipped
    ]),
    AlwaysRunAction, # also skipped with skip_all_remaining!
  ]
end
```

In this case, `AlwaysRunAction` will be skipped if `skip_all_remaining!` was called inside the `reduce_if` block. If `skip_remaining!` had been used instead, `AlwaysRunAction` would still execute because the "skip" state is normally reset when exiting a scoped block.

# Testing

- [x] Added test scenarios for the new functionality.
- [x] Verified that previous tests still work.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,6 +19,7 @@ LightService is a powerful and flexible service skeleton framework with an empha
 - [Stopping the Series of Actions](#stopping-the-series-of-actions)
   - [Failing the Context](#failing-the-context)
   - [Skipping the rest of the actions](#skipping-the-rest-of-the-actions)
+  - [Skipping ALL the rest of the actions](#skipping-all-the-rest-of-the-actions)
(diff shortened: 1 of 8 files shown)


sncalvo:
This PR already updates the README. Do you need me to document this differently?

adomokos:
Thank you! 💯 🏅

PR #284, opened 2026-02-25
Add reduce_while orchestrator construct

kyle-aungst:
## What

Adds `reduce_while`, a new orchestrator construct that checks a condition before each individual action in its step list, breaking out early when the condition becomes false.

## Why

(I think) there's a gap between `reduce_if` and `reduce_until`:

- `reduce_if` checks a condition **once** and runs all steps (or none)
- `reduce_until` is a **loop** that runs all steps as a batch, then checks the condition

Neither supports the case where you want to run a sequence of actions but bail out between any two of them if a condition is no longer met. Today the only workaround is either chaining multiple `reduce_if` calls with the same predicate (verbose), or coupling the condition check into every action via `skip_remaining!` (breaks separation of concerns).

`reduce_while` would be a more convenient way to accomplish this. It makes a single pass through its steps, evaluating the predicate before each one:

```ruby
def self.actions
  [
    add_to_context(errors: []),
    reduce_while(->(ctx) { ctx.errors.blank? }, [
      FetchModel,
      ValidatePermissions,
      ValidateValues,
      UpdateModel
    ]),
    CreateResponse
  ]
end
```

If we ever run into any `errors`, the remaining scoped steps are skipped. `CreateResponse` still runs because `reduce_while` only breaks its own scope.

## Implementation

- `ReduceWhile` class in `lib/light-service/organizer/reduce_while.rb` — extends `ScopedReducable`, iterates steps individually with `scoped_reduce` per action
- Delegate method added to `Organizer::ClassMethods`
- Not added to the deprecated `Orchestrator`

## Tests

- Acceptance spec with 7 examples: condition gating, mid-list break, condition-false-from-start, failed context, skipped context, `organized_by` tracking, and scope isolation
- ContextFactory spec with 3 examples for truthy/falsey paths
- Test double `ReduceWhileOrganizer` added
- `reduce_while.rb` has 100% line coverage; overall suite coverage unchanged

Made with [Cursor](https://cursor.com)


adomokos:
This looks interesting. Mind fixing the build?

kyle-aungst:
sorry! on-call and got wrapped up in it, thanks for the support!

adomokos:
No worries, released it already! Thanks for the PR!

